PR #28, opened 2026-07-23
fix(controller): increase TTS fetch timeout to 60s and retry packet send on reconnect

kopiro:
### Summary

Fixes two audio streaming issues observed when synthesizing long responses (such as bedtime stories) on ESPHome voice satellites:

1. **TTS Audio Fetch Timeout**:
   - `_fetch_tts_audio` in `em_esphome.py` previously had a hardcoded `ClientTimeout(total=10)` (10 seconds).
   - When generating long speech responses (e.g. 500+ token bedtime stories via cloud TTS like Gemini or local TTS engines), HA's `/api/tts_proxy/` audio generation takes longer than 10 seconds, causing `_fetch_tts_audio` to time out and abort playback.
   - **Fix**: Increased `ClientTimeout` to 60 seconds (`total=60`).

2. **Mid-Stream Reconnect Audio Truncation**:
   - In `em_controller.py`, when a brief Wi-Fi or WebSocket drop occurred during audio streaming (`self.data_ws is None`), `send_data` immediately logged `No data connection` and discarded all remaining audio packets.
   - When the device reconnected a few seconds later, the remaining audio had already been discarded, truncating the playback prematurely.
   - **Fix**: Added a brief retry loop in `send_data` (up to 5 seconds) waiting for `self.data_ws` to reconnect before discarding packets.

### DIFF
--- a/controller/em_controller.py
+++ b/controller/em_controller.py
@@ -275,6 +275,11 @@ class Device:
 
     async def send_data(self, data: bytes):
         if self.data_ws is None:
+            for _ in range(50):
+                if self.data_ws is not None:
+                    break
+                await asyncio.sleep(0.1)
+        if self.data_ws is None:
             log.warning(f"[{self.device_id}] No data connection")
             return
         try:
--- a/controller/em_controller.py
+++ b/controller/em_controller.py
@@ -275,7 +275,7 @@ class Device:
 
     async def send_data(self, data: bytes):
         if self.data_ws is None:
-            for _ in range(50):
+            for _ in range(150):
                 if self.data_ws is not None:
                     break
                 await asyncio.sleep(0.1)
--- a/controller/em_esphome.py
+++ b/controller/em_esphome.py
@@ -1089,7 +1089,7 @@ async def _fetch_tts_audio(url: str) -> bytes:
     for attempt in range(2):
         try:
             async with aiohttp.ClientSession() as session:
-                async with session.get(url, timeout=aiohttp.ClientTimeout(total=10)) as resp:
+                async with session.get(url, timeout=aiohttp.ClientTimeout(total=60)) as resp:
                     resp.raise_for_status()
                     audio_bytes = await resp.read()
             break


wilbowes:
Thanks for this — really appreciate the report, and we've reviewed and confirmed both issues are real.

**On the TTS timeout:** merged the substance of this in #29. You're right that 10s is too tight.

Worth noting we've been quietly hitting it too. `_fetch_tts_audio` retries once, so a first-attempt timeout gets absorbed by a successful second attempt and shows up as a latency spike rather than a failure. Across 198 turns with timing data, two took over 10s between the URL arriving and the decoded audio being in hand — 10.3s and 17.5s — both recorded as successes. The 17.5s one lines up almost exactly with a 10s timeout plus the 0.5s backoff plus a successful retry. Your report is what made me go and look properly.

For reference, our split is `tts_gen` (STT result → URL) at median 2.8s / max 6.3s, and fetch+decode (URL → audio in hand) at median 2.0s / max 17.5s. Fetch time scales with payload — roughly 207KB→1.3s vs 461KB→2.7s — which is far slower than a LAN transfer, so generation is clearly happening during the GET here as well.

This is also landing alongside some ring work that tightens the LED dead-man timeouts, and the thinking spinner spans HA think time *and* the TTS fetch — so I've coupled those two values with a test to stop them drifting apart.

**On the reconnect retry:** I'd like to take this, but not in its current shape. The wait sits inside `send_data`, which is called once per 4096-byte audio chunk — a ~3s reply is around 70 chunks — so it runs per chunk rather than once per drop. In the case you're targeting (data plane blips, control plane survives) it works fine, because the same `Device` object is still in the registry and `data_ws` comes back. But if both planes drop — a reboot or power loss — `handle_control` builds a fresh `Device`, so the old object's `data_ws` never returns and the stale task waits 15s × ~70 chunks, roughly 17 minutes, holding a decoded PCM buffer. Repeated flaps stack those up.

Also a small one: the description says "up to 5 seconds" but `range(150)` × 0.1s is 15.

If you're up for it, moving the wait out of the per-chunk path — one bounded wait at the point the connection drops, then either resume or abandon the stream cleanly — would make this mergeable. Happy to take it as an update to this PR.

Thinking broadly, the timeout is a band-aid. The real fix is to stop buffering the whole response — `em_player.py` already streams a URL through ffmpeg into the same speaker plane with a streaming EQ, and pointing that at TTS would remove the deadline entirely and start playback seconds sooner, since today every reply waits for generate + download + decode + EQ before a single sample plays. That's on my list.

wilbowes:
Hi @kopiro — picking this up properly, sorry it sat.

**The TTS timeout half has already landed.** The 10s cap is now 60s, credited to you in the comment above it. Worth noting it has since been largely superseded by your *own* streaming work in #47 — voice-turn TTS no longer buffers the response at all, so a slow generator is now bounded by `sock_read` rather than a total deadline, which handles the long-bedtime-story case more directly than a bigger number could. The 60s cap still covers standalone announcements.

**The reconnect half I've reworked rather than merged as-is** — see #60, with you as co-author.

The problem is real and worth fixing: we've measured RTT excursions peaking around 1.7s on our fleet, so a blip mid-stream genuinely happens and losing the rest of a response is a poor outcome. But `send_data` is called once per audio period, so waiting inside it means the wait is *per frame*: a device that is genuinely gone would stall every remaining frame in turn, and a stream that should abort in seconds would instead drain for hours holding the voice lock. That failure mode is worse than the truncation it fixes.

So the grace is now a budget for the whole **stream** rather than per frame — a blip costs a pause, a real disconnect costs it once and then fails fast. Also armed on the music feed, which had the same problem mid-song.

One small thing worth mentioning: the description says "up to 5 seconds" but the code was `range(150)` at 0.1s, so 15. I went with 3, on the grounds that this covers a reconnect rather than an outage and the device already buffers ~5.5s.

Closing this in favour of #60. Thanks for both of these — between this and the streaming work you've moved the audio path along considerably.

wilbowes:
Merged as #60 with you as co-author. Closing this one — both halves of what you reported are now in `main`.

Thanks again.

PR #36, opened 2026-07-26
Checkers / Echo Show 5 support

imduffy15:
Completely vibe coded with minimal review other than adding a device and testing that the functionality works.

Opening this PR to share whatever learnings were discovered but I'm OK with this not going anywhere as it is completely vibe coded.

### DIFF
--- a/controller/device_payloads/start_server_checkers.sh
+++ b/controller/device_payloads/start_server_checkers.sh
@@ -1,0 +1,124 @@
+#!/system/bin/sh
+# EchoMuse start script for checkers (Echo Show 5 gen 1, LineageOS 18.1).
+#
+# Differences from the biscuit script:
+#   - No Fire OS services to wait on or kill. There is no echoaudioservice,
+#     no ledcontroller and no "mixer". Critically, LineageOS's audioserver is
+#     left RUNNING: system_server blocks forever on media.audio_policy without
+#     it and Watchdog then reboots the device. See internal/profile.
+#   - No tinymix block. Mixer init lives in internal/profile so it stays in
+#     step with the card/device constants instead of drifting in a shell script.
+#   - No LED handling: this device has a screen, not a ring.
+#
+# Retry policy is unchanged: an exit inside MIN_RUNTIME counts as a failed
+# start, and after MAX_ATTEMPTS consecutive fast exits the inactive A/B slot is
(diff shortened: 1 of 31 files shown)


wilbowes:
@imduffy15 — an update, and an honest scope question rather than a review.

## Something your PR settled for us

I just closed #109, which resolved the Dot's button inputs by name instead of by `event` number. It was written on the theory that inserting a jack could shift the enumeration; hardware testing killed that theory (it doesn't shift), so the PR lost its reason to exist on Dot hardware.

Your profile work is what shows the principle was right anyway, just aimed at the wrong device. On checkers, `event2` is the **touchscreen** while volume is `event6` — so anyone who assumed the Dot's numbering would open the touchscreen and get a silent failure: `evdev.Open` succeeds, nothing errors, the buttons are simply dead. That's a much better argument for not trusting event numbers than the one #109 was making, and it came from you having the hardware.

If we do go further with multi-device support, the natural home is naming the devices in the profile (`mtk-kpd`, `keys`) rather than numbering them. Recording that here so it isn't lost.

## The scope question, which is Wil's to answer and currently open

Being straight with you rather than leaving a large PR sitting: **the direction isn't decided yet**, and it isn't about code quality.

- **The target OS is undecided.** Checkers could end up running FireOS or LineageOS, and that choice changes a great deal about what a device profile even means. Building the abstraction before that's settled risks abstracting over the wrong seam.
- **The touchscreen is a real question, not a detail.** Supporting a screen is a different product with a different surface area — UI, input, power, what a "device" is in the dashboard. Wil isn't sure he wants to take that on.
- **There's a case for staying narrow.** EchoMuse might be better as a voice assistant daemon — mic, speaker, LEDs, buttons, and a satellite protocol — rather than a general repurposing platform for Amazon hardware. That's a smaller thing to keep working, and keeping it working on the Dot fleet is already a lot.

None of that is a no, and none of it is a criticism of the work — the profile abstraction itself is a sensible shape for the problem. But it would be worse to merge it, imply support, and then not maintain it than to be clear that the decision hasn't been made.

So: leaving this open, not stalled by neglect. If you have a view on the FireOS/LineageOS question, or on how much of checkers is useful *without* the screen, that would genuinely help — a voice-only checkers profile is a much smaller question than a full one, and might be the version that fits.

PR #46, opened 2026-07-30
wakeword: run the models on the device, via a dlopen'd ONNX Runtime

wilbowes:
Two commits. **`dcd7cee`** adds the ONNX Runtime binding; **`f4479c7`** adds the means to test it on hardware.

## 1. The binding — `device/internal/wakeword/ort`

Fills in the `Inferer` interface #45 was written against, so the device can score audio instead of only buffering it.

**The runtime is dlopen'd, not linked.** Only ONNX Runtime's MIT C header is vendored; the 12.3MB library ships separately. This is a reliability choice: linked properly, a device missing the library would fail to **exec**, and on hardware whose entire recovery story is "count fast exits, flip the A/B slot", turning an absent optional file into a boot failure is a bad trade. Opened by hand, a missing library is an error return and the device keeps using controller-side wake word.

Checked rather than asserted — the ARM test binary:

```
DT_NEEDED: libdl.so, liblog.so, libc.so     # no libonnxruntime
undefined Ort* symbols: none
dlopen/dlsym: imported
```

`DefaultOptions` bakes in what was measured on Lounge: **one thread, XNNPACK, `allow_spinning=0`**. None is ORT's default and two look wrong without the numbers — spinning alone costs 243% of a core against 36%, and *more* threads lowers latency while raising total CPU, the wrong trade for work that is duty-cycled with 51ms of headroom in an 80ms budget.

(ORT's C API is a struct of function pointers and cgo cannot call one held in a C struct, so the `shim.h` layer is unavoidable however the library is bound.)

## 2. Testing it on the device — `device/tools/oww_probe`

"Correct on the host" and "correct on the device" are two questions, and separating them makes the first cheap. The probe answers *does the binding work on this hardware at all* with **no firmware change, no OTA, and nothing that alters fleet behaviour** — a standalone ARM diagnostic in the same vein as `capture_mics`, safe to run on a live device.

- **Phase 1, correctness:** runs the *same* comparison the host test runs, against the same golden capture of Python's output — melspec, embeddings and scores per chunk, plus a probe block for the classifier.
- **Phase 2, cost:** paces frames at the real 12.5Hz duty cycle and reports process CPU from `getrusage`. This is the number worth having, and it catches what a C benchmark structurally cannot — cgo call overhead and Go GC pressure from the buffering.

Exit status is 0 only if every stage matched Python. Flags for `-threads`/`-xnnpack`/`-spinning` so the measured optimum can be re-derived on hardware rather than trusted.

`fixture.Verify` moved into `internal/wakeword/fixture`, shared by probe and test. The two asking *slightly different* questions is the failure mode that matters — it is the probe's answer that gets trusted, since it runs on the hardware — so identical strictness is now structural. Also drops ~150 lines from `ort_test.go`, which had its own parser and tolerance policy.

`controller/tools/push_file.py` puts an arbitrary file on a device (`ota.py` only writes the firmware slot). Four device traps baked in, each of which produced a convincing wrong answer before: the shell plane dies after ~50s under load (reconnects every 30s); the destination is **deleted** rather than appended to, because resuming on size alone appends to a different file and then reports success; `busybox truncate` silently no-ops (trims with `dd`); and markers are split in the source, since the PTY echoes commands back and an unsplit marker matches its own echo. Success means md5 agreement, nothing less.

## Test evidence

| check | result |
|---|---|
| `go test ./internal/... ./pkg/...` (host) | all pass |
| `go vet ./internal/... ./pkg/...` | clean |
| `pytest controller/tests/` | 118 passed |
| ORT suite vs onnxruntime 1.23.2 | all pass; **skips cleanly** with no library, so CI needs no 12MB download |
| cross-compile `android/arm` armv7 API 22 | builds in the existing compiler image, **no Dockerfile change** |
| `oww_probe` on the host | exact agreement with Python; 1.7ms/frame at 2.5% of one core |

That last row is a deliberate cross-check: 1.7ms/2.5% independently reproduces the previously measured controller cost of 1.43ms/1.8%, which is what makes the measurement believable rather than merely available.

## What is NOT verified

**Nothing has run on a device yet** — no device was touched in this PR. The probe exists precisely so that step is cheap and reversible, and it is the next thing to do. Beyond it: wiring into the mic pipeline, shadow-mode score comparison against the controller, an install path for the runtime, and energy-gating inference so it does not run in a silent room.

Also untested: sustained and thermal behaviour (the probe's phase 2 is minutes, not hours), and RSS over time — weak `__cxa_thread_atexit_impl` resolving to NULL means C++ `thread_local` destructors never run, which is fine if ORT's pool is created once and a slow leak if threads churn.


wilbowes:
## Verified on hardware (Lounge, ONNX Runtime 1.19.2, XNNPACK active)

```
== phase 1: does it reproduce Python? ==
  melspec    ok      worst 5.34e-05 abs, 6.26e-07 of tensor scale
  embedding  ok      worst 6.68e-05 abs, 1.37e-06 of tensor scale
  score      ok      worst 8.94e-08 abs (tolerance 1e-05)
  probe      ok      classifier 0.009288 matches Python

== phase 2: cost at the real 12.5Hz duty cycle, 30s ==
  frames     375 over 30.0s (0 overran their 80ms slot)
  latency    p50 31.4ms  p95 32.8ms  max 63.7ms  (budget 80ms)
  CPU        37.7% of one core  (11.32s of CPU in 30.0s wall)
  RSS        22508 KB peak

VERDICT: pass
```

**The headline number is 37.7% against the 36.2% measured in pure C on this same device.** The entire Go binding — cgo call overhead, GC pressure from the buffering — costs about **1.5 percentage points**. That could not be known by reasoning about it, and is the reason the probe was built before any firmware work.

The models on the device are byte-identical (md5) to the ones that generated the fixture, so the residual differences are the engine, nothing else.

## The first run failed, and the criterion was at fault

10 of 24 embedding chunks fell outside tolerance, worst 6.7e-05 absolute. I want to be explicit that this was a real failure and that changing a tolerance after a failure deserves justification rather than a shrug:

- **Per-element relative error is the wrong criterion** for a tensor whose values straddle zero. The same 2e-5 absolute error reads as 1.7e-4 on an element at 0.11 and would read as 2000 on an element at 1e-8 — it fails on whichever element happens to sit nearest zero, while saying nothing about whether the engines agree.
- Against the **tensor's** scale (embeddings span ±45–62 here) the same error is **1.4e-06** — the ~7 significant figures measured independently between ARM and x86 the day before, so two unrelated measurements agree.
- **What settles it is the score.** If the embeddings were wrong in any way that mattered, the one value with a decision attached would have moved. It agreed to **8.9e-08** against a threshold of 0.5.

Tolerances now scale with the tensor, and the probe **prints the agreement, not just a verdict** — "passed at 1.4e-06 of scale" and "passed at 9e-05" tell you different things about a device.

Immediately afterwards the same lesson recurred in miniature: quoting an of-scale ratio for the *score* printed `0.75 of tensor scale` for a run that agreed to 9e-8, because a score of ~1e-7 has no magnitude to be relative to. Alarming, meaningless, and left in a log it would have cost somebody an hour. The ratio is now shown only for tensors with a scale to speak of.

## `push_file.py` got it wrong first too

The first version sent 24KB of base64 per shell command and **wrote nothing at all while every step reported success** — the shell is a PTY in canonical mode, where a line beyond ~4096 bytes is mangled by the line discipline, and the mangling looks like corruption rather than a too-long line.

The fix was to stop inventing and use what already worked: `em_api._stream_file_to_device` moves 10MB of firmware by streaming **short 76-char base64 lines inside a heredoc**, precisely because of this limit. One heredoc per 96KB block keeps resumability; since the plane drops *mid*-block as well as between blocks, each block retries and recovery trims back to the block's start offset with `dd` rather than trusting a partial base64 decode.

Nothing corrupt shipped because success is defined as **md5 agreement and nothing weaker** — that check caught the first failure; the transfer itself never noticed.

## Still not verified

- **Does it hear the wake word in a real room?** Nothing here touches the microphone. That is shadow mode, and it is the next piece.
- **Sustained and thermal behaviour** — phase 2 is 30 seconds, not hours. RSS 22.5MB is a single sample, and worth watching specifically because weak `__cxa_thread_atexit_impl` resolving to NULL means C++ `thread_local` destructors never run.
- **Contention with the real pipeline** — the probe ran alongside whatever Lounge was doing, but not alongside an active voice turn with AEC engaged.

Left on Lounge at `/data/local/tmp`: `oww_probe`, `ort_fixture.bin`, `probe.log` (~2.7MB, plus the library and models already there from the feasibility work). Harmless on 349MB free, and useful for iterating — say the word and I'll clear them.

wilbowes:
## Shadow mode (`712b015`)

The Echo now runs the wake model over its own mic stream and reports what it *would* have detected, acting on nothing. `owwOnDevice` = `off` (default) or `shadow`; schema v13.

### One rule drove the design

**A shadow feature must never be able to damage the real audio path.** Inference costs ~31ms per 80ms frame, the mic goroutine reads 160ms ALSA batches, and the capture ring is only 160ms deep — scoring two frames inline would spend 62ms of that budget and invite the capture stalls that lose whole batches (the failure that silently disabled AEC for four releases). So `Push` hands off to a buffered channel and returns; the scorer goroutine **drops frames and counts them** when it falls behind. A run that drops frames still tells you something. One that stutters the microphone tells you nothing and costs a working device. Inference errors are counted and survived for the same reason.

The tap sits exactly where the ungated wake stream writes to the wire, so the device scores **byte-identical frames on identical 80ms boundaries** — a score difference can then only be the engine, not the framing.

### Nothing per frame

| what | when | cost |
|---|---|---|
| threshold crossing | immediately | rare — a refractory period collapses each utterance to one |
| frames / drops / crossings / max score / errors | existing 30s stats tick | shares the executor hop that write already takes: **one extra upsert per 30s** |

Crossings go immediately because their whole value is the timing; a report arriving 30s late cannot be matched to anything. Unguarded, one utterance emits a burst of crossings, since scores stay above threshold for several consecutive frames.

### Correlation

In `em_shadow.py` rather than `em_controller`, because it is the only genuinely subtle part and `em_controller` cannot be imported by the test suite without dragging in openwakeword and aiohttp.

- **The device never sends a timestamp.** An Echo's wall clock is bogus before NTP, so it reports how long *ago* a crossing happened and the controller converts against its own monotonic clock — same reasoning as the RTT instrumentation.
- **The match window is deliberately loose (2s).** Both detectors see the same frames but not in the same detector *state* — the controller drops wake frames while a turn or TTS is in flight. Of the two possible errors, a false "miss" is the more misleading, because it argues against a feature that is actually working.
- **A match is consumed**, so two turns in quick succession cannot both be credited to one crossing — the discipline `playback_stats` already follows.
- Correlation runs at **turn-persist** time, not detection: the report can land after the wake it belongs to, and by turn end it has had seconds to arrive.

### The distinction that needed most care is in the schema

**A NULL device score means three different things:** shadow was off, the firmware predates it, or the device genuinely did not hear the wake word. Only the last is a finding; the first two are absence of data, and reporting them together would manufacture evidence against on-device detection.

Hence `turns.dev_shadow` — a "the device was known to be scoring" flag that makes a NULL beside it a real miss. `wake_counters.dev_*` carries the hourly view, where crossings with no matching turn are the false-accept side that per-turn rows structurally cannot show. That one is exposed as an **estimate**, not a count, because turn rows and hourly counters are pruned on different schedules and the difference drifts over a long window.

`GET /api/devices/{id}/activity` now returns a `shadow` block with the verdict computed (agreement %, signed delta p50/p95, unmatched crossings, drops) rather than leaving a reader to derive it from raw columns.

### Cost, stated plainly

**~38% of one core, permanently**, on top of the ~18-20% mic-pipeline baseline, on hardware that sometimes has only one core online. Default off; **enable one device at a time**. The runtime and models are installed out of band (`push_file.py`), not shipped in the firmware — 12.3MB would double both the OTA payload and the space each A/B slot takes — and their absence is an ordinary logged condition leaving controller-side wake word exactly as it was.

### Tests

| check | result |
|---|---|
| `pytest controller/tests/` | **136 passed** (12 new correlation tests, 6 new schema tests) |
| `go test -race ./internal/wakeword/shadow/` | pass |
| `go test ./internal/... ./pkg/...` | all pass |
| `go vet` | clean |
| `go build -tags server ./...` | device binary builds |

The race detector was worth running: it caught my tests reading detector state the scorer goroutine owns, and revealed there was no safe way to observe readiness at all (now `Ready()`, which also distinguishes "quiet room" from "never warmed up" in a log). It also exposed a test that pushed 16 frames in a tight loop into an 8-frame queue and silently dropped half.

The scorer's properties — never blocks, drops rather than stalls, one crossing per utterance, survives inference errors — are all tested with a fake `Inferer`, needing no model, runtime, or device. That is the payoff of keeping inference behind an interface.

### Not included

Letting the device *trigger* turns (an unknown `owwOnDevice` normalises to `off` rather than guessing between "score silently" and "start triggering" on firmware that cannot honour it); a dashboard install action for the runtime; and gating inference on acoustic energy to cut the idle-room cost.

**Nothing has been enabled on a device.** Shadow mode is off by default and no config was written — that needs your go-ahead, and Lounge's runtime currently sits in `/data/local/tmp` rather than the canonical `/data/local/share/echomuse/oww`.

wilbowes:
## Deployed to Lounge — and it broke wake word fleet-wide first

Recording the failure before the results, because it is the more useful half.

### What broke

`em_controller.py` does not import `time` — it uses the event loop's clock throughout — and I added `device.last_wake_mono = time.monotonic()` to the wake-detection branch. `NameError`, in an asyncio task whose exception was never retrieved, **on the one line that only executes when someone actually says the wake word**.

So each device worked normally until its first detection, logged `Wake word detected (score=1.000)`, and then silently stopped listening for the rest of the process:

```
07:16:37  [Lounge] Wake word detected (score=1.000, threshold=0.500)   <- listener dies here
07:28:17  [Office] Wake word detected (score=0.980, threshold=0.500)   <- and here
```

No traceback, no "Voice turn starting", no `voice_lock` message. Detection logged, then nothing. Rolled back, fleet recovered, confirmed by Wil on both devices.

**Every test passed, and would pass again.** The wake loop needs openwakeword and a real utterance, so the controller's heavy modules are deliberately not imported by the suite — every rarely-taken branch in them is untested by construction. That is the actual lesson: not "I made a typo" but "this codebase has a large class of code that tests structurally cannot reach."

### Three fixes, in increasing order of importance

1. **The clock lives in one place** (`em_shadow.now()`). Both sides of the correlation are compared by subtraction, so two callers independently reaching for "a monotonic clock" is a latent bug — and a quiet one, since CPython's `loop.time()` happens to *be* `time.monotonic()` today, meaning a mismatch would produce plausible deltas right up until it didn't. `record_cross`'s injection parameter is renamed `at_now` so it cannot shadow it.

2. **`shadow.ModelStem` mirrored `prediction_key` wrongly.** It used `filepath.Ext`, which reads the `.1` of `hey_mycroft_v0.1` as a file extension and sent the device looking for `hey_mycroft_v0.onnx`. A version suffix is not an extension. Now a table test over both forms. This is why shadow never started on Lounge — it failed *closed*, so it was invisible next to the NameError.

3. **CI gains a pyflakes job scoped to undefined names.** This is the systemic fix: it finds the class of bug in milliseconds without importing anything, which is exactly what the suite cannot do here. Verified both ways — passes on this commit, **fails on the line that shipped**. Scoped to undefined names only, deliberately: the codebase has pre-existing unused-variable warnings not worth a churn commit, and a lint job that fails for style gets ignored or deleted.

## Then it worked

Lounge on `v2.9.8-14-g122028b`, controller redeployed, schema v13.

```
[shadow] on-device wake word scoring (onnxruntime 1.19.2, model hey_mycroft_v0.1,
         xnnpack=true, threshold 0.50) — reporting only, not triggering
```

Live comparisons, device vs controller on the same utterances:

```
controller 0.5218  vs device 0.5063  (+185ms)
controller 1.0     vs device 1.0     (-432ms)
controller 0.9885  vs device 0.9885  (+1ms)
controller 0.675   vs device MISS
```

`GET /api/devices/{id}/activity` → `shadow`:

| metric | value |
|---|---|
| turns compared | 5 |
| agreed / missed | **4 / 1 → 80%** |
| delta p50 / p95 | **12ms / 185ms** |
| device score avg / min | 0.796 / 0.506 |
| crossings (unmatched) | 7 (3) |
| frames scored | 2872, **0 drops** |

Two things stand out. The agreeing scores are **near-identical** (0.9885 vs 0.9885, 1.0 vs 1.0), which is the strongest evidence yet that the device pipeline reproduces the controller's. And the deltas go **both ways** (+185, -432, +1) — I predicted device-first, but the controller drops wake frames during turns and TTS, so the two feature rings genuinely lead each other depending on state. The loose 2s match window earned itself.

### Cost, measured rather than predicted

```
Lounge  cpu=51.0%  mem=208MB  shadow=frames=374 drops=0 ready=True
Office  cpu=24.5%  mem=179MB  shadow=off
Retreat cpu=18.4%  mem=188MB  shadow=off
```

**~27-30 percentage points**, not the ~38 I predicted, plus ~20-30MB RSS for the ORT sessions. 374 frames per 30s window is 12.47/sec — exactly the 12.5Hz duty cycle.

And the property the design was built around held: **`stalls=0, sub_drops=0`** on the mic clock after shadow started, with `drops=0` on the scorer. The audio path is untouched.

### Still open

- The one MISS (controller 0.675, device never crossed 0.5) — a real finding, needs more samples before it means anything.
- 3 unmatched crossings — either false accepts or wakes during an active turn the controller ignored. The estimate cannot currently tell those apart.
- Sustained behaviour and RSS drift over hours (the `thread_local`-destructor risk) — 20 minutes is not evidence.

wilbowes:
## `d2f1b07` — thermals, core count, and a metric that was lying

### The Dot has four cores

I reported shadow mode costing "51% of one core, and only one core is online". `/proc/cpuinfo` does show one processor — but `/sys/devices/system/cpu/present` is `0-3`. The MT8163 is a **quad-core** A53 and MediaTek's hotplug strategy parks three of them when idle. A power state, not a ceiling.

`/proc/hps/` governs it: `up_threshold=80` with `up_times=2`, `down_threshold=70` with `down_times=20`, `rush_boost` at 98, `input_boost_cpu_num=2`. cpu0 already runs at its maximum 1.3GHz under `interactive`, so nothing was withheld on frequency either. Lounge sat at 54% — *below* the trigger — so it stayed on one core **because the work fit**, with three cores available on demand the whole time.

### Core floor raised to 2

`num_base_perf_serv` is HPS's core-count **floor** (the `num_limit_*` files are ceilings). Set to 2, it holds two cores online and leaves `up_threshold` to scale to 3 and 4 unchanged.

This is for **deadlines, not throughput**: the mic pipeline has a hard 160ms budget (the ALSA ring's whole depth) and now timeshares with inference running in ~31ms bursts. Two cores let those genuinely run in parallel rather than depending on hotplug reacting to a burst already underway. Measured cost: **+0.3°C at the PMIC, no change at the CPU zone**, on a mains-powered device idling at 33°C.

It lives in the firmware (`applyCoreFloor`) because procfs doesn't survive a reboot, and it sets the floor rather than writing `cpu1/online` directly — HPS re-parks that within `down_times`, leaving a setting that looks applied and silently isn't.

### The instructive part

Lounge went **51% → 25.5%**, level with the untouched devices — **with the load unchanged**.

`cpuPct` is derived from the aggregate `/proc/stat` line, so it is a share of *online* capacity. Doubling the cores halves the reported percentage for identical work. Read carelessly that says "the core floor fixed our CPU problem", which is satisfying and false. `coresOnline`/`coresTotal` now ship beside it — status page and hourly rollup — so a `cpu_avg` series can no longer show a drop that is only a change of divisor.

### Thermals (schema v14)

11 zones: `mtktscpu` (CPU), `mtktspmic` (PMIC), `tmp103` (discrete board sensor), plus board-temp siblings. `maxTempC` is the hottest of all of them, since trouble need not appear on the zone you thought to watch. Amber at 70°C, red at 85°C.

**`thermalCoreLimit` (`num_limit_thermal`) is the sharper instrument than any temperature** — below `coresTotal` means the governor is *already* capping capacity, which bites well before a reading looks alarming. Renders as "throttled 3/4".

An unreadable sensor stores **NULL, never 0**: a zero averaged into a mean reads as a *cool* device, the wrong direction for a metric whose only job is to warn. There's a test for exactly that, and one that older firmware sending no thermal fields reads as absent rather than "0°C, 0 cores".

## Field results: music playback with shadow on

The load case I expected to be the problem. Music Assistant FLAC flow to Lounge, shadow on, two cores:

| signal | during music | idle + shadow | Office (no shadow) |
|---|---|---|---|
| CPU | **26.0%** | 25.5% | 25.0% |
| shadow frames / drops per 30s | 374 / **0** | 374 / 0 | — |
| mic stalls / sub_drops | **0 / 0** | 0 / 0 | — |
| CPU / PMIC temp | 31.5 / 33.2°C | 33.0 / 34.2°C | — |
| thermal core limit | 4 (no cap) | 4 | — |

Cumulative: **34,348 frames, 0 drops**. Music costs essentially nothing on top of shadow — the decode, EQ and pacing are controller-side and the device only writes ALSA. My expectation that sustained playback would be the breaking case was wrong.

Two things attributed honestly rather than left looking clean:

- The single speaker underrun at 08:44:16 was **my own controller restart** at 08:44:15. Music started at 08:44:50, so it cannot be the load.
- The mic clock deficit stepped once by exactly one 160ms batch (−460 → −620) then went flat, oscillating −608/−620 over four further samples. The sign matters: audio is *ahead* of wall, so this is ALSA-vs-wall skew — present at −461ms hours before shadow existed — not lost audio.

### Verdict

Sustainable. Real work is ~0.5 core either way; two cores gives ~75% headroom on the constrained resource, no hotplug wait, thermals 30°C below anything interesting, zero drops or stalls under load.

The only risk left is the one that can't be shortcut: **RSS drift** (+0.46MB/h over 36 min, against a pre-existing ~1.2MB/h baseline — inconclusive) and the weak `__cxa_thread_atexit_impl` meaning C++ `thread_local` destructors never run. That needs hours.

Next optimisation, ahead of anything else on-device: **energy-gate the inference**. A quiet room is most of the day and the mic path already computes RMS.

PR #47, opened 2026-07-30
controller: stream Home Assistant TTS during generation

kopiro:
## Summary

- Pipe Home Assistant's TTS response into one long-lived ffmpeg decoder and yield 48 kHz mono PCM as it becomes available.
- Keep one stateful `StreamingEQ` instance and one device speaker session alive for the complete response.
- Preserve cancellation, a single end-of-stream marker, playback metrics, and the existing buffer-drain behavior.
- Keep standalone ESPHome announcements on their existing buffered path.

## Root cause

Voice turns fetched the complete `/api/tts_proxy/` response with `resp.read()` before starting ffmpeg, EQ, or speaker playback. Streaming-capable TTS providers therefore provided no latency benefit: EchoMuse waited for every utterance to be synthesized and buffered before sending the first PCM period.

This is the structural follow-up suggested in #28. It removes the full-response fetch deadline from the voice-turn path instead of extending that deadline again.

## Impact

Playback can begin as soon as ffmpeg decodes the first audio. Encoded and decoded responses are no longer accumulated in memory, and a retry is attempted only before PCM has been emitted so already-heard audio cannot repeat.

The reconnect retry from #28 is intentionally not included. Text-originated pipelines without an active microphone turn are also outside this PR; this change is limited to native ESPHome voice turns.

## Validation

- `python3 -m py_compile controller/em_controller.py controller/em_esphome.py`
- Live ESPHome satellite validation: the TTS request returned successfully, the first PCM period was sent through the streaming path, and playback completed without a stream error.

### DIFF
--- a/controller/em_controller.py
+++ b/controller/em_controller.py
@@ -573,6 +573,13 @@ class Device:
         pending = bytearray()
         total_pcm = 0
         eq_seconds = 0.0
+        # Accumulated time spent WRITING to the socket, excluding time waiting
+        # for the source to produce audio. send_ms is documented as socket-write
+        # time that "completes near-instantly however slow the link is" — timing
+        # the whole streaming loop instead would fold HA's synthesis time into
+        # it and make it read like delivery, which is the misreading that cost
+        # an investigation on 2026-07-20.
+        send_seconds = 0.0
         first_send_time = None
         try:
             async for pcm in pcm_chunks:
@@ -595,7 +602,9 @@ class Device:
                             f"[{self.device_id}] First streamed PCM period "
                             "sent to device"
                         )
+                    _t_send = asyncio.get_event_loop().time()
                     await self.send_data(bytes([SPEAKER_FRAME_TYPE]) + chunk)
+                    send_seconds += asyncio.get_event_loop().time() - _t_send
 
             if pending and not self.cancel_event.is_set():
                 chunk = bytes(pending)
@@ -607,7 +616,9 @@ class Device:
                         f"[{self.device_id}] First streamed PCM period "
                         "sent to device"
                     )
+                _t_send = asyncio.get_event_loop().time()
                 await self.send_data(bytes([SPEAKER_FRAME_TYPE]) + chunk)
+                send_seconds += asyncio.get_event_loop().time() - _t_send
         finally:
             self.speaking = False
             # One EOS terminates the complete response. Sending EOS per HTTP
@@ -617,7 +628,7 @@ class Device:
             except BaseException:
                 pass
 
-        return total_pcm, int(eq_seconds * 1000), first_send_time
+        return total_pcm, int(eq_seconds * 1000), first_send_time, int(send_seconds * 1000)
 
 
 # The live device registry — keyed by device_id (ro.serialno).
@@ -1049,7 +1060,12 @@ async def _run_streaming_post_turn_playback(device: Device, pcm_chunks) -> int:
         device.eq_bands,
         device.eq_loudness,
     )
+    # Cleared BEFORE streaming starts: the device sets it when its audio
+    # channel drains after EOS, and a stale set from the previous response
+    # would end this turn the moment we started waiting.
+    device.playback_done.clear()
     cancel_task = asyncio.create_task(device.cancel_event.wait())
+    done_task   = asyncio.create_task(device.playback_done.wait())
     stream_task = asyncio.create_task(
         device.stream_speaker_chunks(pcm_chunks, stream_eq)
     )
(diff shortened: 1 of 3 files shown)


wilbowes (commented):
Reviewed against the voice path's existing contracts. **The approach is right and the implementation is careful** — this is the structural fix #28 asked for, not another deadline extension. Before merging, one defect in persisted metrics and two robustness items.

### What's right, and deliberately so

- **One `StreamingEQ` for the whole response.** Per-chunk `apply()` would click at every boundary — the same reason `em_player` carries filter state. Correct, and the comment says why.
- **One EOS for the whole response.** Per-chunk EOS would make the device repeatedly prime and flush, and would disarm the flush-discard contract mid-response. Also correct.
- **Retry only before PCM has been emitted.** Retrying after would repeat audio the user already heard. That's the right rule and it's enforced by the `emitted` flag rather than assumed.
- **Announcements left on the buffered path.** Right call — they have no streaming benefit and a different lifecycle.
- `playback_send_t0` now anchors on the *actual* first period rather than the pre-stream timestamp, which makes `delivery_ms` more accurate than it was.

### Blocking: two persisted metrics silently change meaning

`send_ms` is documented in CLAUDE.md as *"a socket-write time and completes near-instantly however slow the link is — never read it as delivery; that mistake cost a whole investigation on 2026-07-20."*

Under this PR it becomes the whole fetch + decode + stream duration:

```python
t_stream_start = loop.time()          # before any audio exists
...
device.playback_send_ms = int((loop.time() - t_stream_start) * 1000)
```

So a column whose entire documented purpose is "this is *not* delivery time" starts measuring something very close to delivery time, in existing rows, with no migration and no rename. The next person to read that history will draw exactly the conclusion the comment warns against.

Same class of problem for `tts_fetch_ms`. `trace.t_tts_fetched_ms` is now stamped *after* `post_turn_play` returns, i.e. after playback and the drain sleep have completed — so `fetch_ms = t_tts_fetched_ms - t_tts_url_ms` now includes the entire playback. It was fetch+decode only, and it's the metric that diagnosed #28 in the first place.

Suggestion: keep `send_ms` as pure socket-write time (measure inside `stream_speaker_chunks` around the `send_data` calls, or from `first_send_time`), and stamp `t_tts_fetched_ms` when the first PCM arrives — for a streaming design "time to first audio" is the meaningful number and it's genuinely better than what it replaces. A new column for total stream duration would be welcome, just not on top of these two.

### Should fix: ffmpeg outlives a barge-in

`_stream_tts_audio_once` cleans up in `finally`, but on cancel the async generator isn't closed explicitly — `stream_task.cancel()` raises inside `async for`, leaving the generator suspended and its `finally` to run whenever the loop's asyncgen hooks get to it. Barge-in is a routine path here, so that's a lingering `ffmpeg` per interruption in the window before GC. An explicit `aclose()` on the generator in `_run_streaming_post_turn_playback`'s `finally` would make the teardown deterministic.

### Should fix: `stderr=PIPE` is never drained until after stdout EOF

```python
while pcm := await proc.stdout.read(16 * 1024):
    yield pcm
await feeder
err = await proc.stderr.read()
```

If ffmpeg ever writes more than a pipe buffer (~64KB) to stderr it blocks, stops producing stdout, and both sides wait forever. `-loglevel error` makes it unlikely — but the case that produces lots of stderr is a malformed or truncated response, which is exactly the degraded case this path now has to survive. Draining stderr concurrently (as `feeder` does for stdin) removes it.

### Question, not an objection: the drain sleep goes to ~zero

`remaining = audio_duration - playback_elapsed`, where `playback_elapsed` is measured from the first period sent. When HA synthesises at roughly realtime — the normal case — `playback_elapsed ≈ audio_duration` and `remaining ≈ 0`, even though the device still holds up to ~5.5s in `audioChanDepth`.

The ring is safe (it waits on `playback_stats`, not this sleep), but the mic restart isn't: the comment at the call site notes the reference implementation's 0.5s settle is "already covered by the buffer drain sleep". With streaming, that cover largely disappears, so the mic may open while the device is still playing — the acoustic-feedback case the barge-in-off branch stops the mic for. Worth deciding explicitly rather than inheriting.

### Testing

Understood that the suite can't reach `em_esphome` (openwakeword/aiohttp), so `py_compile` plus a live turn is a reasonable bar. If you want the metric semantics pinned cheaply, they're assertable at source level — `tests/test_deploy.py` already does that style for other invariants.

Happy to take any of these on if you'd rather not — particularly the metric question, since the constraint comes from our history rather than anything visible in the code.

wilbowes:
Pushed the review fixes to this branch (thanks for enabling maintainer edits) — original commit untouched, mine on top with you as co-author. Happy to drop them if you'd rather take a different approach to any of it.

**One was more serious than I first filed it.** I raised the drain sleep as a *question*; it's actually a regression. Streamed playback ended on a computed `audio_duration - playback_elapsed` sleep, and that estimate was removed from the buffered path on 2026-07-24 — it has no visibility of the device's own buffer and cleared the ring **6.1s early on Retreat, 3.2s on Lounge**. Streaming makes it worse rather than better: the streaming loop already spans most of the audio duration, so the remainder computes to ~0 and the wait disappears entirely while up to ~5.5s is still queued in `audioChanDepth`. Restored the `playback_done` wait with the same timeout backstop and cancel race the buffered path uses.

Nothing you could reasonably have known — it's a six-day-old fix whose reasoning lives in a comment on the function you replaced rather than the one you wrote.

**The rest, as reviewed:**
- `send_ms` back to socket-write time only, accumulated around the `send_data` calls. It had become fetch+decode+stream, i.e. very nearly delivery — the exact reading CLAUDE.md warns against.
- `t_tts_fetched_ms` now stamps the **first** decoded audio instead of the end of playback. Time-to-first-audio is both truthful and more useful for a streaming design than what it replaced.
- The async generator is `aclose()`d explicitly — its `finally` is what kills ffmpeg, and barge-in is routine here, so leaving it to GC keeps a decoder alive.
- ffmpeg's stderr drained concurrently rather than after stdout EOF, so a full pipe can't deadlock decoder and reader.

Two source-level tests pin the parts that would regress silently (the `playback_done` wait, and `send_ms` staying write-time). 158 controller tests pass; the suite still can't reach this path directly, so a live turn remains the real check.

**Verdict from my side: this is the right change** — it deletes the `resp.read()` that made streaming providers pointless and implements what the comment in `_fetch_tts_audio` had been asking for, reusing the pattern `em_player` already proved. Not a band-aid. Merging is @wilbowes' call, pending a live voice turn.

Issue #53, opened 2026-07-31
Music Playblack Pause Issue


nabarge:
I love EchoMuse!
I had been hanging on to a few of these echo dot 2's since I saw that they had been rooted hoping something like this would be developed. So first off, THANK YOU!

Two pause related issues.

If I ask the echo to play music, it works fine, but after a few moments the esphome media player reports that it is idle (the icon shows the play arrow rather than the pause icon) even though the music continues to play on the echo. Kinda weird and causes some inconsistent state issues. 

Secondly, when I "Barge-in" the music gets paused until the interaction is over, then un-paused. This generally works great, unless I want to actually pause the music. If I try to pause the music (lets say from music assistant) I get the response that music is already paused, then the interaction ends and after a few moments it un-pauses the music. I can't actually get the music to stop and stay stopped via voice alone, I have to open up the music assistant web/app interface and manually press pause.

(This also gave me an idea. Can the action button be configured in Home Assistant. Like give options for when it does, or enable long/double/triple press to run actions or automations.)

Thanks for all your efforts, and being so generous with you work!

wilbowes:
Hey, thanks for the feedback. I also noticed the playback issue yesterday, then got sidetracked, so thank you for reminding me so I can look into it. 

I can look at customising the action button, would be a nice feature. I'd probably keep control of the single press for starting a voice turn and cancelling an ongoing turn. Double or triple tap I might be able to detect and then pass through as events to HA for you to do what you want with.

wilbowes:
Both of these are fixed, and released in **v2.9.13** (device) + **controller-v2.12.0**. Thanks for the detailed report — the second one in particular was more interesting than it looked.

**The media player reporting idle while music plays** turned out to be two separate causes wearing the same symptom. The feed announces `playing` exactly once, when the decoder starts, so anything sent afterwards becomes the last word Home Assistant hears — and two places were asserting a hardcoded `IDLE` regardless of what was actually happening: the end of every voice turn, and every volume report from the device. So "after a few moments" was the turn ending, and nudging the volume would do it too. Both now report the real state, and there's a test that forbids the shape rather than just those two instances.

**Not being able to pause the music** was worse than it appeared. Barging in pauses the music for the turn, and `pause()` returns early unless something is playing — so your pause was **silently discarded**, you were told it was already paused, and the auto-resume then contradicted you at the end. Tellingly, "stop" worked, because that path already cleared the resume flag.

[...]

PR #55, opened 2026-08-01
player: let a user's pause during a barge-in win over our auto-resume

wilbowes:
Fixes half of #53.

Barging in pauses the music for the turn. If the thing you then ask for **is** "pause the music", it failed in about the most confusing way available: `MediaSession.pause()` returns early unless the state is `PLAYING`, so the command was **silently discarded**, the user was told the music was already paused, and `resume_interrupted` then started it playing again at the end of the turn.

A pause arriving while *we* hold the session paused is the user taking ownership of that state, so the auto-resume goes.

**`stop()` has always cleared that flag** — which is precisely why "stop the music" worked during a barge-in and "pause the music" did not. This makes pause agree with stop rather than adding a new rule. Cleared at the module-level entry point, not in `MediaSession.pause()`, because `interrupt()` calls that method itself right after setting the flag.

Three tests: the reported sequence (verified by reverting the fix and watching it fail with the user's exact symptom), the stop path it is now consistent with, and the ordinary barge-in that must still resume on its own. 219 controller tests pass.

**On ducking instead of pausing** (@wilbowes' suggestion) — not attempted here, and worth knowing why before anyone tries it: the device buffers ~5.5s ahead, so ducking applied controller-side would not be audible for several seconds. It has to happen **device-side at the ALSA write**, where the meter already taps RMS — new control message plus firmware. That is a feature, not a fix, and this bug should not wait for it.

Also spotted while in here, not changed: `resume()` during an interruption starts the feed **immediately**, so a "play the music" mid-turn would collide with the TTS. Narrower and odder than the reported bug, so flagging rather than folding in.

Still open in #53: the media_player entity reporting idle while audio continues.


wilbowes:
**Scope grew, deliberately** — second commit turns this from a pause fix into the ownership model it should have been.

Patching `pause()` alone left the same bug reachable two other ways. `interrupt()` only paused what was **already** playing; nothing stopped a command from **starting** playback mid-turn, and `em_esphome` fires all four playback commands with a bare `create_task`. So `resume()` mid-turn started the feed straight into the TTS — and so did `play()`, which is the more common path, because "play some jazz" runs the intent *before* HA generates the spoken reply.

`interrupted` was a fact about the past ("we paused this"). It is now two facts about the present: `owned_by_turn` and `resume_after`. While a turn owns the speaker, `play`/`resume`/`pause`/`stop` record intent instead of touching the wire; releasing applies it, **last write wins** ("play jazz… actually, pause"). A user command **overrides** the auto-resume — theirs is an instruction, ours is bookkeeping — which is what makes the original #53 fix an ordinary case of the rule rather than a special case bolted onto `pause()`.

Ownership is taken **unconditionally**, including when nothing is playing. That is exactly the case the old code ignored and the case "play some jazz" hits.

Deferred commands still push the **intended** state to HA, so the entity is not stale for the length of a turn — #53's other half is already a complaint about that entity being wrong.

**16 player tests, 220 controller tests. Four of the new tests fail if the deferral is removed — verified by removing it.** Deployed and running; 3 devices connected, no errors.

Still open in #53: the media_player reporting idle while audio continues. Not touched here.

PR #73, opened 2026-08-02
Home Assistant app (previously add-on)

Pinball3D:
This PR brings the ability to run the controller software as a Home Assistant app (previously add-on) There are still a couple things that need to be fixed but given that this is still a beta software, and it is vibe-coded, I'm sure they will get fixed.


wilbowes:
Closing this in favour of #122, which has now merged — but the idea and several of its mechanics came from here, and that's worth recording rather than glossing over.

@Pinball3D got there first and worked out the two things that made the approach viable: bridging `/data/options.json` into the env vars the controller already reads, so `em_controller.py` never has to know Home Assistant exists, and making the dashboard's paths ingress-relative. #122 credits this PR explicitly for both. It was a fresh implementation against current `main` rather than a fix on top of this branch mostly for mechanical reasons — this one was 124 commits behind with a conflict — not because the thinking was wrong.

For what it's worth, "vibe-coded" is underselling it. Getting an add-on to work behind Supervisor's ingress proxy is fiddly in ways that aren't obvious until you try, and having a working example to review made the second attempt much faster than a first attempt would have been.

Thanks for it, and for the gracious note on #122. If you fancy another go at something, the auth passthrough question you raised there is still open and it's a more interesting problem than it looks — HA has already authenticated the user by the time ingress forwards a request, but removing our own login makes the ingress source-IP check the entirety of the add-on's auth, and that check is itself unverified on real hardware so far.

PR #107, opened 2026-08-08
Add buttonSingleTapEvent and buttonMultiTapMs: tap as an HA event

bboe:
Two opt-in settings, both off by default. With neither set, behaviour is unchanged.

**buttonSingleTapEvent** — a tap fires `single` on the HA event entity instead of
starting a voice turn. Trades cancel-by-tap for a button HA can bind to. Gated on
`button_hold`: without that capability HA was never offered the event entity, so
the setting is refused rather than leaving the button inert.

**buttonMultiTapMs** — taps inside the window coalesce and report once as
single/double/triple. Requires buttonSingleTapEvent, since it delays every tap.
0 disables. The window lives in `em_tap_burst.py` with its own tests, same split
as `em_arbiter.py`.

Also taken from review: the Advanced section copy, and a line in the toggle help
steering destructive automations to `long`.

Rebased onto current main, so the tap branch sits before #108's mute check — a tap
in event mode fires while muted, same reason a hold does. No conflict for you.

Left as follow-ups, both device-side: a ring flash on tap, and the stale volume arc.

Ran on one Dot 2 (FireOS 5.5.5.4) with a 300ms window, on the pre-review build —
the capability gate and the coalescer extraction are tested but not yet on my
hardware.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -161,9 +161,15 @@ directions.
   reported on release), never by timing the down/up messages controller-side:
   RTT excursions past 1600ms have been measured on this fleet, which would
   turn a 750ms gesture into noise. Absent `heldMs` reads as a tap, so old
-  firmware keeps its existing behaviour. Only `long` is emitted — double and
-  triple were parked deliberately, because knowing a press was *single*
-  requires delaying every single press by the multi-tap window.
+  firmware keeps its existing behaviour. A hold always fires `long`; a tap
+  fires `single` only under `buttonSingleTapEvent`, which makes the tap an HA
+  event rather than a voice turn. Double and triple were parked deliberately,
+  because knowing a press was *single* requires delaying every single press by
+  the multi-tap window — a tax on the primary action, which is why they are
+  only available once a tap is no longer the start of speech.
+  **`buttonSingleTapEvent` is gated on `button_hold`** — the event entity is
+  only advertised for a hold-capable device, so on older firmware the setting
+  is refused rather than leaving the button inert.
   **Mute blocks the voice TURN, not the gesture.** A hold fires `long` while
   muted; only the tap-starts-a-turn path is refused (`em_button.decide`, with
   the mute state read off the press itself — the device sends `muted` on every
@@ -178,6 +184,8 @@ directions.
   mute, not the button filter. A controller with a stale mute view can at
   worst start a turn that captures silence and ends `no_speech` — keep that
   rejection in `cmd/server.go` where it is, it is what makes this safe.
+  A tap under `buttonSingleTapEvent` fires while muted too, for the hold's
+  reason: it is an event, not speech.
   **The evdev reader must filter to `EV_KEY`**: every press is followed by an
   `EV_SYN` whose code and value are both 0, and without the filter that SYN
   read as a release microseconds after the press. The button therefore acted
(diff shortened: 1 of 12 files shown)


wilbowes:
Thanks — this is well put together, and the reasoning for why the multi-tap
delay is acceptable here is the part I'd have wanted argued. Parking
double/triple was about not taxing the primary action; once a tap is an event
rather than the start of speech, that tax is gone. Agreed, and it's the reason
to take this rather than do it differently.

Checked out and ran the controller suite: 294 passed. The config plumbing is
complete in both directions (including the `dashboard.jsx` `CONFIG_SECTIONS`
mirror and the Dockerfile `COPY`, both of which get missed), the
`BUTTON_EVENT_TYPES` comment about `event_types` being one-shot at connect is
the right call, and re-checking `enabled()` at window expiry rather than at tap
time is a detail I'd have expected to have to ask for.

(That 294 is against your base, which is a few commits behind `main` — current
`main` is at 311. No conflicts reported, but worth pulling it forward before
the final push.)

One change needed before I merge.

## Gate both settings on the `button_hold` capability

The event entity is only advertised when the device declares `button_hold`
(`em_esphome.py:431`). That capability landed 2026-08-01 and ships in v2.10.0
only — every earlier release, v2.9.13 down, doesn't have it.

On such a device with `buttonSingleTapEvent` on:

- no `heldMs`, so every press reads as a tap and takes the new branch
- the branch sends `single`, but the entity was never advertised, so there's
  nothing in HA for it to reach
- the voice-turn and cancel branches are now unreachable

The button ends up completely inert — no turn, no cancel, no event, and no
`long` either. The only trace is the new `dropped — HA not attached` warning,
which points at the wrong cause.

Two parts to the fix. Controller-side, don't take the tap-event branch unless
the device is `button_hold` capable, so old firmware keeps its current
behaviour. Dashboard-side, disable the toggle with the reason rather than
letting it silently do nothing — same pattern as `shadowCapable` at
`dashboard.jsx:1735` and the `mixCapable` prop beside it. That's the
degrade-to-old-behaviour rule in CLAUDE.md's compatibility section.

My own fleet is on v2.10.0, so I wouldn't have hit this; anyone who hasn't
taken the latest firmware would.

## Not blocking, but worth deciding on

**Mute disables the button — but that one's ours, not yours.** The device drops
dot presses at source while the mic is muted, so even the `long` we shipped in
v2.10.0 never fires with the mic off. Pre-existing, nothing to do with this PR;
your change only widens who notices, since a tap is a more likely binding than
a hold. Fixed separately in #108, which moves the rule to the controller as
*"while muted, don't start a voice turn"* — so gestures that aren't turns get
through.

That has a bearing on the ordering here. Once #108 lands, `handle_button_event`
has a mute check between the hold branch and the voice-turn branches, and your
tap-event branch belongs **before** it: with `buttonSingleTapEvent` on a tap is
an event rather than speech, so it should fire while muted for exactly the same
reason a hold does. Whichever of us merges second gets the conflict — happy for
it to be me if you'd rather not rebase.

**A tap has no local feedback in this mode.** Today a tap paints the listening
ring. In event mode nothing happens on the device at all, so there's no way to
tell a registered tap from a missed one — which also makes the two failure
modes above hard to diagnose from the device. A brief ring flash on tap would
cover it. Happy for that to be a follow-up.

**Volume arc.** The device calls `CancelVolumeDisplay()` on every dot release
(`server.go:169`) to hand the ring over to the listening frame, and
deliberately doesn't repaint, because the controller's frame lands within an
RTT. In event mode that frame never arrives, so adjusting volume and then
tapping clears the arc early and leaves the ring on stale `baseLEDs`. Cosmetic,
and it'll get reported as a ring bug rather than a button one.

**Section copy.** The Advanced section's description still says "Everything
here affects only bounded button-press turns". With `buttonSingleTapEvent` on,
the setting abolishes button turns rather than tuning them. One clause.

## One thing for people reading this later

Worth a line in the toggle's help text steering destructive automations to
`long`. Before this, the only bindable gesture was a hold, which is hard to
trigger by accident. A tap on a speaker sitting on a counter is not, and the
button carries no authentication — so "unlock the door" bound to `single` is a
different risk class from the same automation bound to `long`.

CI hasn't run yet — GitHub is holding the workflow for approval since it's your
first PR here. I'll approve it alongside the next push.

No need for a JOURNAL.md entry, and you're right about why.

bboe:
Thanks for the prompt review @wilbowes!  The capability gate is added, and the em_button refactor landing
first made it a cleaner fix than it would have been.

I've rebased onto current main. The tap outcome is now part of `em_button.decide`,
ordered before the mute check as your comment there suggested, gated on
`button_hold` and ANDed by the caller. The dashboard toggle disables with the
reason. Also took the section copy and added a line steering destructive
automations to `long`.

Multi-tap's window moved out to `em_tap_burst.py` with its own tests; the
`em_arbiter` split, for the same reason.

Happy to leave the ring flash and the volume arc as follow-ups.

Running on my Dot 2 since the rebuild. Muted taps can't be tested here yet as
this Dot is on v2.10.0 and b55bbce isn't in a release yet, so the press never
leaves the device.

wilbowes:
Merged, thanks @bboe — this is a good change and it came back better than it went out.

Verified before merging: the capability gate is where it needs to be (`TAP_EVENT` in `em_button.decide`, ordered before the mute check, ANDed with `button_hold_capable` at the caller), so a pre-v2.10.0 device keeps its existing behaviour rather than ending up inert. Dashboard toggle disables with the reason, following the `shadowCapable`/`mixCapable` pattern.

The rebase onto current `main` picked up #108, so the merge-order conflict I flagged resolved itself in your favour — moving the tap outcome into `decide` was the right call and made that a non-event.

333 tests green locally and all five CI jobs green. Pulling `em_tap_burst` out into its own module with its own tests was worth doing.

Ring flash and volume arc left as follow-ups as agreed. On muted taps: you will be able to test those once b55bbce is in a release — until then the press never leaves the device, so there is nothing wrong with your setup.

wilbowes:
Follow-up now this is running on our hardware, @bboe — and the headline is that it works.

**`single` and `long` both come through to Home Assistant correctly.** Tap fires the event instead of starting a voice turn, hold still fires `long`, and the ordering you moved into `em_button.decide` behaves.

**`double` is unreliable here, and it is not your code.** It fired once and then proved hard to reproduce at a 150ms window. Measuring rather than guessing turned up the cause: `TapCoalescer` restarts the window when a tap *arrives at the controller*, so the gap it measures is the real gap plus the difference in RTT between the two taps. On this fleet that error term is large:

| probes | avg | min | max | over 200ms |
|---|---|---|---|---|
| 2566 (Test Device, recent windows) | 26–470ms by hour | 1ms | 9255ms | **26.4%** |

A quarter of control-plane probes exceed 200ms, and the spread inside a single window runs from single-digit milliseconds to seconds. So a genuine 120ms double-tap frequently lands more than 150ms apart and reads as two singles. It can compress as well as stretch, so a slow-then-fast pair can merge taps that were not a burst.

This is the same trap the hold path already dodges — `heldMs` is measured on the device precisely because RTT excursions past 1600ms have been measured here. Nothing about that is documented anywhere you would have seen while writing this, and your own Dot may well sit on a much quieter link, so I would not have expected it to show up in your testing.

Worth being clear that the coalescer itself is right: restarting the window per tap, re-checking `enabled()` at expiry rather than at tap time, and resetting the count before emitting are all correct, and the reasoning in the module docstring about why the delay is acceptable still holds. The timing *source* is the only issue.

Tracked as #115. The fix is to have the device report the gap since the previous tap from its own monotonic clock, alongside `heldMs`, and let the controller use that to decide burst membership — the expiry timer can stay where it is, since it only has to answer "has the burst stopped". That needs a firmware field, so it will ride the next firmware release rather than a controller change.

In the meantime **`buttonMultiTapMs: 350` works well here for both double and triple** — tested on hardware, not estimated. Human double-taps land roughly 150–400ms apart, so 150ms was tight even before jitter entered into it. Worth considering as the default rather than 0 for anyone enabling multi-tap.

None of this changes the merge — `single` is the case the feature exists for and it works. Thanks again.

bboe:
Thanks for merging! Is it okay if I let you make the adjustments for defaults when the feature is enabled? 0 as the default simply means multi-tap is not enabled, but moving the default to 350 is also completely reasonable and I think you know what's best for the product.

Issue #80, opened 2026-08-03
Support for external audio jack


wilbowes:
External audio needs a detection mechanism to switch away from the internal speaker. Adding as an enhancement.

nabarge:
I did notice this issue today, trying the external jack. It seems like if I boot the speaker with an aux cable already plugged in, the speaker doesn't boot quite right, and I cant even get the wakeword detection to work and can't even get the buttons to work. If I unplug and replug the cable, or mute and unmute ... or press some magic button combo I could get it to play audio over the internal speaker but not through an external output. 

If I booted the speakers without the aux cable in, and then plug an aux cable in after it was fully booted, I could get sound to switch to an external speaker, (although sometimes it seemed to "freeze up" or pause the music...but not every time). When I unplugged the aux cable it would not switch back to the internal speaker until I rebooted the speaker.

brmo:
Well that makes sense!  I just tore down my dot 2 as I am making a new enclosure for the dot 2 with a new speaker with an amplifier. I destroyed the 3.5mm jack so I can solder the wires directly to the board and its having the same issue as above.

wilbowes:
Went and looked at a live device. The detection mechanism this issue says
does not exist does exist, and there is a likely explanation for @nabarge's
buttons dying.

## Detection is already there

```
/sys/class/switch/h2w/state        -> 0
/proc/bus/input/devices            -> N: Name="ACCDET"   (event0)
/sys/bus/platform/devices/soc:accdet@/modalias
                                   -> mediatek,mt8163-accdet
```

`h2w` is Android's standard headset switch: 0 nothing, 1 headset with mic, 2
headphone. The MediaTek accessory-detect driver is bound and running, and it
also publishes an input device.

Same shape as the ambient light sensor: invisible from the layer we normally
look at, sitting right there on the raw interface. So detection is a file
read, and this issue is really about **routing**, which is the part still
unknown.

## Why the buttons might die when booting with a cable in

@nabarge: booting with the aux cable in left you with no wake word and no
buttons. There is a plausible cause, and it is ours.

The button reader hardcodes its input devices by index:

```go
const dotButton    = "/dev/input/event1"
const volumeButton = "/dev/input/event2"
```

On a device with nothing plugged in, the enumeration is:

| event | name |
|---|---|
| event0 | ACCDET |
| event1 | mtk-kpd (action + mute) |
| event2 | keys (volume) |

Nothing guarantees that order. If accdet probes differently with a jack
[...]

PR #109, opened 2026-08-08
Resolve button input devices by name, not event number

wilbowes:
Item 1 of the three in #80. Independent of the routing question, and worth
doing on its own merits.

## The bug

The evdev reader hardcoded its inputs by index:

```go
const dotButton    = "/dev/input/event1"
const volumeButton = "/dev/input/event2"
```

On a Dot with nothing in the jack the enumeration is ACCDET, `mtk-kpd`,
`keys`, so those constants are correct. Nothing guarantees that order. The
MediaTek accessory-detect driver publishes ACCDET and probes ahead of both
keypads, so if it enumerates differently with a jack already inserted,
everything after it shifts by one and the reader opens the wrong device.

**The failure is silent.** `evdev.Open` succeeds on any input node, so the
outcome is working hardware, no error anywhere, and a dead action button.
That matches @nabarge's report in #80: booting with an aux cable in left a
device with no buttons and no wake word, recoverable by unplugging and
replugging.

## What this does not claim

It does not confirm the numbering shifts with a cable in. Nobody has booted a
device that way and dumped `/proc/bus/input/devices`, so that remains a
hypothesis that fits the symptom. This makes the question moot rather than
answering it.

## Behaviour

Names are stable, event numbers are an enumeration accident — the same rule
the ambient light sensor already follows (`0-0039`) and the thermal zones
(resolved by type).

Resolution **falls back to the historical paths** when the name is not found,
so an unreadable procfs or a differently-named keypad on another SKU behaves
exactly as it does today rather than losing its buttons. Every outcome is
logged with what *was* present, so an absence is diagnosable from a log line
instead of a round trip asking someone to run commands on a device they have
just flashed.

On current hardware this resolves to `event1`/`event2` — i.e. it is a no-op
until the day it isn't.

## Tests

6, all pure. The fixture is verbatim from Office including the trailing
whitespace the device actually prints, since a tidied fixture would not catch
a parser that trips on it. Covered: the real output, a shifted enumeration,
the event token being found rather than indexed (`Handlers` carries
`gpufreq_ib` and `dynamic_boost` around it), a nameless stanza not inheriting
the previous name, empty input, and the fallback.

Cross-compiles clean (`./compile.sh`). Not yet run on hardware — on a device
with no jack it should log `"mtk-kpd" resolved to /dev/input/event1` and
behave identically.

### DIFF
--- a/device/internal/bindings/buttons/evdev_controller.go
+++ b/device/internal/bindings/buttons/evdev_controller.go
@@ -9,6 +9,8 @@ import (
 	"os/exec"
 )
 
+// Historical paths, now only the fallback when resolution by name fails.
+// See resolve.go for why the numbering cannot be trusted.
 const dotButton = "/dev/input/event1"
 const volumeButton = "/dev/input/event2"
 
@@ -49,11 +51,11 @@ func (e *EvDevController) SubscribeToButton(callback buttons.ButtonClickCallback
 
 	dotBtn := e.GetDotButton()
 	volBtn := e.GetVolumeButton()
-	dotDevice, err := evdev.Open(dotButton)
+	dotDevice, err := evdev.Open(resolveInputDevice(keypadDeviceName, dotButton))
 	if err != nil {
 		return nil, err
 	}
-	volDevice, err := evdev.Open(volumeButton)
+	volDevice, err := evdev.Open(resolveInputDevice(volumeDeviceName, volumeButton))
 	if err != nil {
 		return nil, err
 	}
--- a/device/internal/bindings/buttons/resolve.go
+++ b/device/internal/bindings/buttons/resolve.go
@@ -1,0 +1,110 @@
+package buttons
+
+import (
+	"log"
+	"os"
+	"strings"
+)
+
+// Input device NAMES, which are stable, rather than event numbers, which are
+// an enumeration accident.
+//
+// On a Dot with nothing in the jack the order is ACCDET, mtk-kpd, keys, so
+// the buttons land on event1 and event2 and the old hardcoded constants were
+// right. Nothing guarantees that. The MediaTek accessory-detect driver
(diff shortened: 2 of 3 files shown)


wilbowes:
Closing this. The reasoning it was built on turned out to be wrong, and hardware testing on the audio jack (#80) is what showed it.

## The premise is falsified

This PR argued that accdet might probe differently with a plug already inserted, shifting every input device after it, and that this explained the dead buttons in #80. Measured on a cold boot **with the plug in**, the enumeration is identical to one without:

```
event0  ACCDET
event1  mtk-kpd     (action + mute)
event2  keys        (volume)
```

So the hardcoded constants are correct on this hardware in both states, and there is no observed case where they are not.

#80 turned out to be something else entirely: Android's mediaserver holds the speaker PCM when a headset is present at boot, so our blocking `snd_pcm_open` never returns and `main()` never reaches `SubscribeToButton`. The buttons were dead because the handler was never registered — the evdev nodes were fine all along. Fixed in #113.

## Why not merge it anyway as housekeeping

Three things which together outweigh the tidiness:

- The doc comment in `resolve.go` explains itself in terms of the jack scenario. Merging would put a **wrong explanation** into the codebase beside the code it justifies, which is worse than not having the code — the next person reasons from it.
- It has never run on hardware, and it touches the most user-visible input path. The realistic outcomes are "no change" or "buttons break".
- #36 is already changing the same two constants, from someone who owns the hardware in question.

## The idea is right, the aim was wrong

Worth preserving, because #36 validates the principle while killing this implementation: on the Echo Show 5, `event2` is the **touchscreen** and volume is `event6`. Open the touchscreen and you get exactly the silent failure this PR was written to prevent — `evdev.Open` succeeds, no error, dead buttons.

So "don't trust event numbers" is correct; it just needed different hardware to matter. The natural home is name resolution inside a per-device profile rather than a global scan — but that depends on decisions not yet made (see #36), so it goes on the pile rather than into main.

Issue #33, opened 2026-07-26
Home Assistant App/Add-on


wilbowes:
Home Assistant native controller and dashboard to remove the need for an external docker instance.

wilbowes:
Shipped. The controller runs as a Home Assistant Supervisor add-on — repository at the root of this repo, add-on config in `controller/config.yaml`, with the dashboard served through ingress in the sidebar.

Some detail that is easy to miss and matters if you install it:

- **Auth is Home Assistant's.** Supervisor forwards the authenticated user and strips any client-supplied copies, so on a genuine ingress request those headers are proof of an HA session. The first HA user through the door becomes admin, which also removes the bootstrap-token step you get on the standalone container.
- **Roles are not mirrored from HA.** Everyone after the first is read-only until promoted. Supervisor forwards no admin flag, and asking HA for one costs a permission far too broad to justify (`auth_api` includes `POST /auth/reset`, which sets any HA user's password with no verification).
- **There are two channels**, GA and Early Access, as separate add-ons with their own slugs — the shape ESPHome uses, since HA has no channel concept.
- **Channels share no storage**, so switching is a migration: a new database *and a newly generated CA*, which strands devices holding the old one. Copy `/data/tls/` first. Same trap applies moving from the standalone container — `docs/migrate-to-addon.md` covers it.

The add-on and the standalone container are both first-class and neither may gain or lose capability relative to the other; #163 tracks the two `.env` variables not yet reconciled.

Closing this as done.

PR #122, opened 2026-08-10
Package the controller as a Home Assistant add-on

natecj:
## Summary

Adds a `config.yaml` + `repository.yaml` so the controller can be installed
as a Home Assistant Supervisor add-on (Settings → Add-ons), instead of
running `docker compose` by hand. Home Assistant itself still needs no
custom integration — this only changes how the controller is deployed.

- `em_start.py` bridges the add-on's `/data/options.json` into the env vars
  `em_controller.py` already reads (`.env.example`), so `em_controller.py`
  itself stays unaware of Home Assistant.
- The dashboard gets ingress support: relative asset/API paths behind a
  `<base href>` the server injects, gated by `ECHOMUSE_HOME_ASSISTANT_INGRESS`
  so non-add-on deployments (docker-compose) are unaffected.
- `host_network: true` (same reason `docker-compose.deploy.yml` uses
  `network_mode: host` — mDNS + per-device ESPHome ports).
- `translations/en.yaml` gives each option a friendly name/description in
  the add-on's config UI.
- `icon.png`/`logo.png`, generated from the existing dashboard favicon, so
  the Store listing isn't a generic placeholder.
- README gets a "My Home Assistant" badge to add the repository in one click.

## Related

#73 attempted the same thing and is where several of the ideas here
(ingress-relative paths, an options.json launcher) came from — credit for
that. This is a fresh implementation against current `main` rather than a
fix on top of that branch, because a few things needed to change:

- #73 is 124 commits behind `main` with an unresolved conflict.
- Its `require_device_tls` option is silently broken: it sends `"true"`/`"false"`
  to `em_controller.py`, which checks `== "1"` — the toggle does nothing.
- Its `arch:` only lists `aarch64`, though the published image is multi-arch.
- Its `repository.yaml` points at a personal fork rather than this repo.
- Its per-option labels/descriptions live in a `configuration:` block inside
  `config.yaml`, which isn't where Home Assistant reads them from (they
  belong in `translations/en.yaml` — confirmed against the official ESPHome
  add-on and verified live against Supervisor's API), so they likely don't
  render at all.

Related Issue: #33 

## Test plan

- [x] `python -m pytest tests/` — 336 passed, including a new
      `test_dashboard_paths_are_ingress_safe`
- [x] `pyflakes em_start.py em_api.py` — no new warnings
- [x] Installed on a real HA OS host: local multi-arch build succeeds,
      options → env mapping verified, dashboard loads and logs in through
      HA's ingress panel (setup page and full dashboard + WebSocket)
- [x] Config labels verified live via Supervisor's API
      (`/addons/<slug>/info` → `translations.en.configuration`)
- [x] Icon/logo verified live via Supervisor's API
      (`/addons/<slug>/info` → `icon: true, logo: true`)
- [x] A real rooted Echo Dot 2nd Gen connected to the add-on-hosted
      controller and completed full voice turns (wake word → STT → HA →
      TTS → playback) end to end


natecj:
@wilbowes I can't assign reviewers so tagging you here.

A related question is about auto-login when running through HA - what are your thoughts on skipping the username/password part in this scenario to reduce friction? If that sounds reasonable I can try working on that as a followup PR.

wilbowes:
Thanks, and apologies I haven't had a chance to look at this yet. The auth passthru (and other questions) is something I've been mulling for a while. I'll have a proper look later today and make some decisions.

Pinball3D:
Thanks for fixing my shit lol! I was just trying to get something scrapped together... @wilbowes on that note about auth, that is something I think can be redone. I don't think it needs to generate a token to create the account. Realistically there is no risk of someone else creating the first account and credentials.

wilbowes:
Reviewed, and I've pushed three fixes plus current `main` straight onto this branch (`maintainer_can_modify`) rather than asking you to round-trip them. Shout if you'd rather they came as a separate PR.

**`oww_threshold: 0.3` → `0.5`.** This one matters more than it looks. #144 changed that default on the strength of 212 fleet-hours of near misses peaking above 0.3, and #155 merged it a few hours ago. A fresh add-on install is the *only* case where a shipped default is visible at all — an existing deployment already has every key stored, so `DEFAULT_DEVICE_CONFIG` is never consulted — which makes `config.yaml`'s copy the one users actually meet. It's now pinned against `em_db` by a test.

**Added `image:`.** Without it Supervisor builds this Dockerfile on the user's Home Assistant machine, which compiles onnxruntime and ffmpeg on what is usually a Pi. It also can't pass `EM_CONTROLLER_VERSION`, so the controller reports `dev` and the update notice goes quiet. `controller-release.yml` already publishes `linux/amd64,linux/arm64` matching your `arch:` list, so pulling works as-is. `version` moves to `2.18.0` because Supervisor pulls `image:version` and that's the tag GHCR actually has.

**`em_start.py` comment vs code.** The `OPTION_ENV_VARS` comment said an unmapped option "fails loudly here"; the code did `continue`. It now logs the drift by name — warning rather than exit, since failing to boot over one stray key is worse than ignoring the setting.

Both new guards were verified by reintroducing the bug.

## Two things still open, and one is yours to answer only after Wil decides

**The ingress gate needs testing on real hardware.** `_ingress_only_middleware` rejects on `request.remote != "172.30.32.2"`. With `host_network: true` the container shares the host's network namespace, so I'm not confident the address Supervisor's ingress presents is that gateway IP. Wrong one way and the dashboard is unreachable; wrong the other and it's on the LAN with nothing in front of it. That can't be settled by reading — it needs an install. Wil is planning to deploy this on his own HA shortly, so it'll get exercised.

**Auth passthrough.** Still Wil's call, and I've put the tradeoff to him: HA has already authenticated the user by the time ingress forwards a request, so a second login is friction — but skipping it makes `request.remote` the entirety of the add-on's auth, which is the same check above. That ordering is why I'd want the gate proven before the login goes away, not alongside it.

Nice work on this — the `em_start.py` boundary and the ingress-relative path handling are both cleaner than they needed to be, and shipping tests with it made this review much quicker.

PR #136, opened 2026-08-11
build: drop esbuild from the runtime image

bboe:
esbuild is a build-time tool — it compiles `dashboard.jsx` and is never executed at runtime — but two copies ship in the published image:

```
/usr/local/bin/esbuild
/tmp/package/bin/esbuild     # extracted tarball, never cleaned
```

The existing `rm` removes only the `.tar.gz`.

Both are Go 1.20.12 binaries, so a vulnerability scan reports **18 HIGH/CRITICAL per copy**, including `CVE-2024-24790` (CRITICAL, `crypto/x509`) and `CVE-2023-45288`. That's 36 findings on binaries the container never runs.

This adds both paths to the `rm` that already runs after the compile. `dashboard.js` is produced exactly as before — verified byte-identical output and a working controller image (`em_controller` / `em_api` / `em_arbiter` all import, dashboard 248KB).

### DIFF
--- a/controller/Dockerfile
+++ b/controller/Dockerfile
@@ -100,7 +100,7 @@ RUN mkdir -p /app/models/oww_runtime && \
 # Dashboard SPA — placeholder is index.html, full dashboard is dashboard.html
 COPY static/ ./static/
 # Compile JSX dashboard with esbuild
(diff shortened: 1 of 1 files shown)


wilbowes:
Merged, thanks — good catch.

Worth recording why the fix is in the same `RUN` and not a later one: a `rm` in a subsequent layer would leave the bytes in the earlier layer, so the image would be no smaller and the scanner would still find both copies. Deleting alongside the compile is what actually removes them.

PR #166, opened 2026-08-13
Support static controller endpoints for routed and tunneled devices

NicFragale:
## Summary

Adds an optional persistent static controller endpoint for EchoMuse devices
that cannot discover their controller through link-local mDNS.

When `/data/local/etc/echomuse/controller.json` exists and is valid, the
device skips mDNS and repeatedly dials the configured controller. Existing
mDNS behavior remains unchanged when the file is absent or invalid.

Example:

```json
{"host":"10.20.40.110","port":8767,"tls_port":8770}

### CODE REVIEW

**Comment by wilbowes** [comment]
**Yes to this, and sorry for the silence. Eleven days is too long for a PR this well formed.**

There is an existing design it belongs to, #106, opened before this one and finished off tonight. That is on me for not linking it sooner.

**You got the hard part right.** #106 exists mainly to flag one trap: `tls_port` is only ever advertised as an mDNS TXT record, so a device configured with a static address *because mDNS does not work* has no way to learn it, dials plain, and then gets refused by a controller running `REQUIRE_DEVICE_TLS=1`. Carrying `tls_port` in the file closes that by construction rather than by recovery. Your parsing, validation and table tests carry over as they are, IPv6 included.

Three changes to fit the rest of it:

1. **A list, not a single endpoint.** A static address, a backup address and a DNS name all behave identically, so one ordered `endpoints` array rather than three keys. Migration then means appending an entry, and the order is what controls a cutover.
2. **Fall through rather than pin.** This treats the static endpoint as authoritative and never reaches mDNS. The overriding goal is that a device is never left unable to reach a controller, so a stale address has to fall through to the next option. A device side `"mdns": false` covers the case pinning is really for, which is keeping a test fleet off the production controller.
3. **Re-read the file each round**, the way `tlscreds.go` re-reads credentials on every dial. Read once at startup means a typo needs a restart to fix, and a device that cannot connect is exactly the one you cannot easily restart.

One thing from the same design that affects this directly: most disconnects should not run the hierarchy at all. A controller restart or a poor link is a reconnect to the same place, and rediscovering on every drop puts multicast browsing on a link that is already the problem. The device pins whatever worked and only starts hunting once that has failed for a while.

Yours if you want it. Say either way and I will pick it up if not.

**Comment by NicFragale** [comment]
@wilbowes I reworked per your notes:

- `controller.json` now takes an ordered `endpoints` array instead of single `host`/`port`/`tls_port` keys; a static address, backup, and DNS name are just entries in the list, tried in order.
- A dial failure falls through to the next endpoint (round-robin) instead of pinning to a stale one forever. Pending-approval doesn't count as a failure; that endpoint's fine, the device just isn't approved yet.
- The config file is re-read at the top of every reconnect attempt, same contract `loadLinkCreds` already uses for the TLS credential files; editing it takes effect on the device's very next retry, no restart.
- Each endpoint still carries its own `tls_port`, so the TLS-discovery gap you flagged (`REQUIRE_DEVICE_TLS=1` refusing a plain-ws static dial) stays closed.
- One more since first pass: a static endpoint now gets 2 consecutive attempts before falling through, so a single transient blip doesn't cost a full lap through the rest of the list before retrying the one actually in front of the device.

Pushed as `933e30e` (on top of `03e2b91`). Docs updated to match (`docs/configuration.md`).

**Comment by wilbowes** [comment]
**Three of the four landed, and the fourth is my wording rather than your reading.** Sorry for the two-day gap — and for a longer one you could not see: CI had never actually run on this PR until today, because it was sitting behind the first-time-contributor approval gate. It is green now, 10/10, and merges clean.

Confirmed in the diff: the ordered `endpoints` array, the per-round re-read, `tls_port` per entry, and pending-approval not counting as a failure. The two-attempts-before-rotating detail is a good addition nobody asked for.

**"Fall through" meant reaching mDNS once the list is exhausted, not rotating within the list.** As it stands, `if len(staticEndpoints) > 0` takes the static branch and mDNS and last-known are both unreachable for the life of the process — the code comment still says "authoritative and bypasses mDNS entirely". Both readings of what I wrote are fair; this is the one the design needs:

```
configured list, in order → persisted last-known → mDNS
```

each falling through, unless the device config carries `"mdns": false`. The reason is that a device must never be left unable to reach a controller, so a list that has gone stale has to have somewhere to go. The case pinning exists for — keeping a test fleet off the production controller — is what `"mdns": false` is for, and TLS verification already stops a device attaching to another fleet's controller if it does fall through.

**Your objection to that is right, and it is the part that needs solving.** `discovery.FindServer` retries indefinitely, so falling into it plainly would park the device there and strand it when a tunnel later comes up. So the mDNS step has to be **bounded** — one attempt inside the cycle, not the open-ended browser. The cadence in the design is one fast pass through every option at ~5s each, a second pass, then backoff 5 → 10 → 20 → 60s, which gives mDNS its turn without letting it swallow the loop.

I have put the settled design at the top of #106 — it was sitting in a comment while the issue body still described an older proposal, which is how this got ambiguous in the first place. That is on us.

**Say the word if you would rather I finish it.** You have reworked this once already and the remaining change is the bounded-mDNS step, which is fiddlier than the rest; it is your PR and your name on it either way. Equally happy to hand it back to you.

Two things from the same design that are **not** yours to solve and shouldn't hold this PR: the persisted last-known file (`last_controller.json`, separate from the file the operator owns), and stickiness — pinning whatever endpoint worked and reconnecting only to that until a grace expires, so a controller restart doesn't run the hierarchy at all. Happy for those to land separately on top of this.

---

**Comment by NicFragale** [comment]
Cool..."fall through" now means what you meant: exhaust the static list, then one bounded mDNS attempt, not indefinite FindServer. Implemented against the settled design in #106:

- controller.json gains "mdns": false for the pinned-test-fleet case, defaulting to true.
- A pass = every static endpoint once, then one FindServerOnce browse if mdns is enabled.
- Two fast passes (~5s/attempt), then backoff 10 → 20 → 60s, holding there.
- errPending counts as success for discovery purposes...it doesn't advance the pass or perturb backoff.
- The disconnected-ring callback now fires once per failed pass rather than once per failed target, so a routine controller restart doesn't flash the ring on every intermediate endpoint.

Left the persisted-last_controller.json tier and stickiness out, per your note.... I can pick either up separately once this is in place.

**Comment by wilbowes** [comment]
**The bounded-mDNS step is right, and the pass structure is better than what I asked for. One blocking bug: the reset you wrote can never run, so the backoff never comes back down.** CI is green now — the run was sitting behind the first-time-contributor gate again, which is on us, not you.

Two things worth calling out first because they are the parts that are easy to get wrong and you didn't. `MDNS *bool` so absent is distinguishable from explicit false, with the reasoning about why the opt-out has to be device-side — a controller-side switch would also kill HA's discovery of the ESPHome and BLE-proxy services on the same stack — is a better argument than the one in #106. And firing the ring callback **once per pass rather than once per target** is exactly right; PR #321 needs that boundary and is currently inventing its own, worse version of it.

## 1. `err == nil` is unreachable, so nothing ever resets the backoff

```go
} else {
    // A clean end to a connection that did establish is not
    // a discovery event — always restart at the top of the
    // list on the next attempt...
    targetIdx, targetAttempts, passNum = 0, 0, 0
}
```

After registration, `connect` sits in a read loop whose only exit is `return err` from `ReadJSON`. There is no path that returns `nil` from an established connection — so this branch is dead, and `passNum` is never reset by anything except a pending-approval.

What that costs: a device that took five passes to find its controller at boot holds `passNum = 5` for the life of the process. Six hours of healthy operation later the link blips, and the first retry waits **60 seconds** instead of 5 — and it re-enters at whatever `targetIdx` it stalled on rather than at the operator's first endpoint, which is the drift your own comment says the reset exists to prevent.

The reset wants to hang off a **successful connection**, not a nil return — next to `c.connectedCallback()` in `connect`, which is where the state is actually known good.

One caveat before you move it there, because a plain reset-on-connect trades one problem for another: a device in a connect-then-drop flap would reset its backoff every cycle and hammer. The codebase already answers this — `WAKE_RESTART_HEALTHY_S` in `cmd/server.go`, "ran at least this long before dying = a fresh incident, not a loop". Same shape, and worth reusing rather than re-deciding.

## 2. The mDNS slot gets two browses per pass, not one

The mDNS target is just `targetIdx == len(static.Endpoints)`, so it goes through the same `targetAttempts >= maxStaticAttempts` gate as a static endpoint and is tried twice. Your commit message says "one bounded mDNS attempt", the code comment says "one bounded mDNS browse", and the log line prints `(%d/%d)` against `maxStaticAttempts` — three statements of a contract the code doesn't keep.

`maxStaticAttempts = 2` is justified by a specific argument — a blip on the endpoint physically in front of the device shouldn't cost a lap through every backup. That argument doesn't transfer to the mDNS slot, which has no "in front of the device" to protect. Suggest the mDNS target advances after one attempt.

## 3. `device/internal/.DS_Store` is in the commit

6148 bytes of Finder metadata. It also isn't in `.gitignore`, so it's worth adding there in the same push.

## 4. A question, not a defect: is the backoff meant to space attempts or passes?

`wait = staticRetryDelay(passNum)` is evaluated after every attempt, so the delay multiplies by the number of targets. With five endpoints plus mDNS at two attempts each, a pass at `passNum >= 4` is twelve dials at 60s — about twelve minutes before the list comes round again, and up to twelve minutes to notice a controller that came back.

Spacing attempts is defensible; I only want to know it was chosen. If it was, a line in the comment saying the pass period scales with list length would stop the next person "fixing" it.

## 5. An invalid config file falls through to mDNS even when `mdns:false` was set

```go
static, staticErr := discovery.ConfiguredEndpoints()
if staticErr != nil {
    log.Printf("[control] Static controller config invalid — using mDNS: %v", staticErr)
}
```

A parse error returns `nil`, so `usingStatic` is false and the device browses — including for a pinned test fleet whose whole reason for `mdns:false` is never reaching the production controller by accident. One typo in the file and it can attach to the wrong one.

Failing open is a reasonable default and I am not sure it is wrong; a device that can never connect is also bad. But it should be a decision in a comment rather than a consequence of where the error is handled. TLS verification limits the blast radius, not the behaviour.

## 6. Nit

`dialErr` from an empty mDNS round surfaces as `Connection lost: mDNS: no controller found this round`. Nothing was lost — it just didn't find anything.

## Where this leaves it

Fix 1, 2 and 3 and I'll merge it. 4 and 5 are answers rather than changes, and either answer is fine.

Worth knowing it has a queue behind it: **#321 (the connecting-ring animation) is blocked on this PR** and I've said so on it. Your `passNum` is the boundary that work needs, so the sooner this lands the better for someone else's PR too.

NicFragale:
Fixed 1–3, and recorded decisions on 4–5 rather than changing behavior (you said either answer works, and I think the current one is right for reasons below). Pushed as b122df2.

1. Backoff reset was dead code. You're right that connect() never returns nil... confirmed, its only exit after registration is ReadJSON's error. Rather than reuse WAKE_RESTART_HEALTHY_S... I couldn't find it anywhere in cmd/server.go or the rest of device/, so either it's from a different branch or the reference didn't make it into what I'm looking at... I implemented the same concept fresh: connect() now returns whether the connection stayed up at least staticHealthyDuration (30s, past one keepalive round trip) before it ended, and Run() resets targetIdx/targetAttempts/passNum on that signal instead of the unreachable err == nil check. A connect-then-drop flap under 30s still backs off normally.

2. mDNS now has its own maxMDNSAttempts = 1, separate from maxStaticAttempts. Agreed it doesn't get the same "don't abandon your primary endpoint" argument.

3. .DS_Store untracked, added to .gitignore.

4. Backoff is per-attempt, deliberately... documented in a comment on staticBackoff. A long list does mean a slower pass at the 60s tier, but the design's own examples are 2-3 entries, and I'd rather keep individual dials evenly paced than add a second timing model for large lists nobody's configuring yet.

5. Invalid config still fails open to mDNS, mdns:false included... also now a comment, not just a consequence of the code shape. A device that can't parse its own file shouldn't be stranded any more than one with no file at all, and TLS verification is the actual backstop against reaching the wrong fleet either way.

6. Fixed: mDNS-miss now logs "no controller found via mDNS this round," not "Connection lost."

wilbowes:
**Merged.** The pass structure, the bounded mDNS slot and the healthy-connection
reset all went in as they stand.

We also fixed the CI gate that kept catching you. It was never your branch:
GitHub treats a contributor as first-time until something of theirs is merged, so
every push you made re-armed the approval wait. That happened three times on this
PR. We have changed the repo to only gate accounts new to GitHub, and this merge
takes you out of it anyway.

Two pieces are still unclaimed: the persisted last-known tier and stickiness.
There is also a third we have just filed as #509. Nothing writes
`controller.json` to a device yet, and the device that needs a static endpoint is
the one we cannot reach over the shell plane to give it one. Say if you want any
of them.

This reaches devices on the next firmware tag, not with the controller, so it
will be a `v*` release and an OTA.

Issue #162, opened 2026-08-13
Audio distorting above certain volume levels


mangowxlf:
When setting volume levels over 80% there is absurd amounts of distortion (via auxiliary out and internal speaker)

This can be partially solved by configuring DSP settings in music assistant to have -13dB of gain reduction on output per each EchoMuse device but doing so doesn't resolve the distortion when voice assistant is speaking.

Personally I suggest adding a volume gain slider within the playback settings per device (maybe -35db to +6db range??) in the echomuse device manager as it may be different per device when/if it will distort above certain volume levels or if a device potentially were to be too quiet at full volume. 

Obvious volume warning on this, just for demonstration of the issue

https://github.com/user-attachments/assets/4e27a580-116d-4bef-9e1d-2e0929a6d61a

wilbowes:
Fixed. Ships in **`controller-v2.20.0`** and firmware **`v2.12.0`** (both are out — update the controller with `docker compose pull && docker compose up -d`, then push firmware from the device's Updates tab).

Your instinct that it was gain, not the speaker, was right. tinymix ctl 61 is the tlv320aic32x4 DAC *digital* volume: 176 steps of 0.5dB spanning −63.5…+24dB, with 0dB at index **127**. We shipped `volumeMax = 175` — the control's own maximum — so the top 27% of the range applied up to +24dB of digital gain to already near-full-scale PCM and saturated inside the DAC. 127/175 is 72.6%, which is where "over 80%" lands once you account for the button steps.

Measured on hardware (1kHz at −6dBFS, recorded back through the mic array):

| index | THD |
|---|---|
| 127 | 1.5% |
| 136 | 2.3% |
| 153 | **65%** |
| 170 | **89%** |

Output level was *flat* from 153 upward — it had stopped being able to get louder — with h3 at −1.1dB relative to the fundamental, which is very nearly a square wave. The control that isolates it: index 170 with the source scaled down to land at the same acoustic level reads 1.1%, i.e. clean. So the gain stage is fine and it was purely source × gain exceeding full scale, which is also why your −13dB of DSP reduction in Music Assistant helped and why it did nothing for the voice response — that path does not go through Music Assistant.

Worth knowing: stock FireOS never writes this control at all. It is absent from `/system/etc/audio_device.xml` and from every binary in `/system`, so the DAC sits at its 0dB reset default and Alexa's volume comes from AudioFlinger's software attenuation, which only ever attenuates. That is why native Alexa has no such distortion and we did.

[...]

Issue #141, opened 2026-08-11
Still getting some strange behavior on aux input.


nabarge:
When the aux cable is plugged it the controller shows  dozens of so disconnects and reconnects. If music is playing the music regularly pauses, and restarts, sometimes doesn't restarts, and even skips forward and backward though the song. If I unplug the cable from the aux port the behavior generally goes back to normal (although sometimes it seems like the echo needs to reboot).

wilbowes:
Yeah, just confirmed it's not quite right. Will investigate further.

wilbowes:
Confirmed, and thanks for the report — it reproduces here. Same underlying fault as #117, where I have posted the technical detail.

**What is happening:** with something in the aux jack, the microphone capture pipeline stalls repeatedly (about every 102 seconds). When a stall runs past 10 seconds the controller's keepalive gives up and drops the connection — and that teardown takes the Home Assistant media player entity with it. The entity dying and rebuilding every couple of minutes is exactly what you are seeing as music pausing, restarting, and jumping around the track. So the disconnects and the music behaviour are one problem, not two.

**Workaround that is better than rebooting:** unplug the aux cable and plug it back in. That clears it immediately — I did it five times tonight and it worked every time. It will come back after a few minutes with the cable in, but it does not need a reboot.

**What we know so far:** it is not the cable. A 3-pole cable, a 4-pole cable and ordinary headphones all trigger it. Every audio register on both chips is identical between working and broken, and the device is demonstrably sending well-formed audio the whole time, so it is not a configuration problem in our software that we can simply correct. The fault is in the analog/jack-detection domain and we have not found the mechanism yet.

Two things are coming regardless:

1. **The disconnect handling is disproportionate** and we are going to fix that (#140). A brief hiccup should not destroy your media player entity. That will make this much less disruptive even before the jack fault itself is understood.
2. The next diagnostic step is comparing against a Dot still running Amazon's stock firmware, which will tell us whether this is our software or the hardware.

If you are able to add a couple of details it would help: what is plugged in (powered speaker, headphones, amplifier?), and does unplug-replug clear it for you the way it does here?

nabarge:
Yes it does seem that unplug-replug clears for me the same as it does for you. 
I had these plugged into powered speakers (I tried an alarm clock and an old iPod dock) I tried with 2 echos, and has the same result on both.

wilbowes:
[...]

Issue #117, opened 2026-08-09
Unplugging the headphone jack stalls the mic ~30s and drops the device


wilbowes:
Pulling a headphone plug out stalls the microphone for roughly 30 seconds. The controller sees no mic frames, gives up on the device, and drops it; the device reconnects on its own about 34 seconds later with the **same pid** — it never restarted.

A user pulling a plug should not lose their device for half a minute.

## Observed

Test Device, firmware `20260809-0800-dev`, jack unplugged at ~07:26:12:

```
07:25:51  OWW: no mic frames for 10s on the continuous wake stream — sending defensive mic_start
07:26:01  OWW: no mic frames for 10s ...
07:26:11  OWW: no mic frames for 30s and defensive mic_start isn't helping — escalating to mic_stop + mic_start
07:26:12  [control] Device disconnected: G090LF1180570SPJ
07:26:12  Control send failed: sent 1011 (internal error) keepalive ping timeout
07:26:12  [data] Data connection closed
07:26:46  [control] Device connected (full caps)
07:26:59  Dot button → voice turn        <- working normally again
```

Mic frames stop about a second after the plug leaves the socket. The controller's escalation (`mic_stop` + `mic_start`) does not help, which suggests the capture is stuck below our control rather than in our stream handling.

Recovery is clean: same pid, no supervisor restart, no slot flip, and the next voice turn played with `underruns=0 minDepth=43`.

## Why it matters more now

It predates the #80 fixes, but was effectively invisible: before them, unplugging left the speaker dead until a reboot, so nobody got as far as noticing a temporary mic stall. Fixing the speaker is what exposed this.

## Where to look

The mic is a **different codec and bus** from the speaker — the four `tlv320aic3101` ADCs on i2c bus 0, captured over `amzn_mt_spi` (card 0 device 24), whereas the speaker is `tlv320aic32x4` on bus 2, device 23. So the jack has no obvious business touching the capture path at all, which is the interesting part.

Leading suspicion is that accdet reconfigures shared codec or clock state on the plug-out interrupt, and the SPI capture does not recover until something times out. Things worth checking, roughly in order of cost:

- `dmesg` across the unplug — the accdet driver is chatty (`accdet_check_status- status switch:[...]`), and the ~30s duration suggests a specific timeout rather than a lost interrupt.
- Whether `[mic] capture stall` appears in the device log at the same moment, which would place it in the ALSA read rather than higher up.
[...]

PR #168, opened 2026-08-13
Native AFE: route device audio through Amazon's own front end

kdkavanagh:
Puts EchoMuse's device audio through Amazon's own audio front end — the ASP pipeline inside the Echo's audio HAL — instead of tinyalsa's direct PCM writes, which bypass it entirely.   I reverse-engineered this after finding EchoMuse's default pipeline to 1) sound quite tin-can-y, even with EQ, and 2) not handle AEC as well as native-alexa while playing spoken word media like podcasts

Opt-in per device, default off, old path untouched and still the default.

Disclosure: Implemented completely with Claude under my supervision, (which seems to match the style for the existing project) - Its thoughts below:

-----

Full spec, phase plan, kill criteria and risks: [`docs/native-afe-migration.md`](docs/native-afe-migration.md). Background on the front end itself: [`docs/alexa-afe.md`](docs/alexa-afe.md).

## What's here

`internal/opensl` (dlopen'd OpenSL ES shim, following the `wakeword/ort` precedent so a device where the load fails falls back rather than failing to boot), `internal/bindings/slmic` + `slspeaker` behind the existing `mic.Microphone` / `speaker.Speaker` interfaces, `device/tools/afe_probe`, and the `native_afe` / `native_afe_backend` capabilities with the bypassed config controls rendered disabled-with-reason.

Capture and playback switch **together**, never independently — ASP takes its far-end reference on the HAL's playback side, so a capture-only conversion produces audio that is beamformed but not echo-cancelled, with no error anywhere. That failure looks like the AFE underperforming rather than like a wiring mistake, which is why it is one switch.

`slspeaker` keeps everything the tinyalsa backend carried: the lead ring, the prime-before-start hold, both music/voice planes, the per-sample duck ramp, and `StreamStats`. Ducking stays device-side for the reason it always was — the controller runs 4.0s ahead of realtime, so audio that has left the controller cannot be ducked by it.

**Opt-in is a marker file**, `/data/local/etc/echomuse/native_afe.enabled`, checked by `start_server.sh`. Not a config push: recovery from a bad AFE path is a flag flip and a restart, and a boot-time choice cannot come from a channel that needs the device to have connected first. Hand-exporting the env var in `start_server.sh` was the obvious alternative and does not survive — `_sync_start_script` re-syncs that file on every OTA.

**Dashboard**: an "Audio pipeline" panel on the device's Config tab, offering EchoMuse's pipeline or Amazon's, writing the marker and rebooting.

**Also included, not AFE-specific:** `Cap volume at the codec's unity gain`. tinymix ctl 61 is the DAC's *digital* volume — unity is index 127 of 175 — so the top 27% of the scale applied up to +24dB to already near-full-scale PCM. Measured on hardware: THD 1.5% at 127, **65% at 153, 89% at 170**, with output level flat above 153 because it had stopped being able to get louder. Stock FireOS never writes this control at all. Independent of the rest of this branch; happy to split it out if preferred.

## Bugs found along the way

Three were found by actually running this on hardware, and two of them are the kind that report success:

- **`slspeaker` never enabled `Ext_Speaker_Amp_Switch`.** `PcmSpeaker.Init()` sets it; `slspeaker.NewSpeaker` had no equivalent and its `Init()` was a no-op nothing called. The amp is a physical mixer control that defaults off and is turned off by accdet on headphone insert. So the path delivered every period with `underruns=0`, healthy `minDepth` and `outcome=ok` — into a dead amp. **Inaudible, with no failure visible in any log or in StreamStats.** Worth knowing generally: the delivery instrumentation all sits upstream of the amp, so when playback is silent, read the mixer state on the device before trusting any of it.
- **`internal/aec.Process()` still ran on the passthrough path**, where its own size guard silently no-ops it forever (`FrameSize` is tuned to the 160ms tinyalsa batch cadence, not slmic's 80ms). Harmless, but an `aecEnabled=true` leftover did nothing.
- **AGC had no such guard**, so it genuinely ran on top of the AFE's own AGC during a lockMic turn — real double-processing, not a no-op.

And one in the dashboard, unrelated to hardware: **`Toggle` did not accept a `disabled` prop at all** — only `Slider` did. Beamforming, Echo cancel and Auto gain therefore rendered greyed with their reason, read as off, and wrote the *opposite* value into config when clicked. That is worse than a control that does nothing, since the stored setting then silently disagreed with what the control showed. Fixed in the handler rather than the styling, with a test pinned against the component (the call sites already looked correct while the bug was live).

## Remaining items

Deferred and filed in the doc, not addressed here: `slspeaker`'s buffering depth against the controller's 4.0s lead (a real regression if the default is ever flipped without it), per-device mic tuning becoming fleet-fixed since `AFE.cfg` is on read-only `/system`, and the LED direction arc, which the AFE gives no per-frame beam number for.

## Verification

- ARM cross-build via `echomuse-compiler` (`-tags server`) — succeeds
- `go test ./internal/... ./pkg/...` — all pass
- `pytest controller/tests/` — 370 pass
- Dashboard node tests + pyflakes undefined-name check — pass
- Live on one Echo Dot Gen 2 (`20260813-0732-dev`), voice turns, music playback and barge-in

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1105,6 +1105,44 @@ Reverting a section **discards** its stored values (`set_device_config_sections`
 
 ### Volume / mute persistence
 
+**The scale stops at the codec's unity gain, and that ceiling is load-bearing.**
+tinymix ctl 61 is the tlv320aic32x4 DAC *digital* volume: 176 steps of 0.5dB
+spanning −63.5…+24dB, with 0dB at index **127**. The firmware shipped
+`volumeMax = 175` — the control's own maximum — so the top 27% of the range
+applied up to +24dB of digital gain to already near-full-scale PCM and
+saturated inside the DAC. Measured on hardware 2026-08-13 (1kHz at −6dBFS,
+recorded through the mic array): THD 1.5% at index 127, 2.3% at 136, **65% at
+153, 89% at 170**, with the output level *flat* from 153 upward because it had
+stopped being able to get louder, and h3 at −1.1dB relative to the fundamental
+(very nearly a square wave). The control that isolates it: index 170 with the
+source scaled down to land at the same acoustic level reads 1.1% — clean — so
+the gain stage is fine and it is purely source × gain exceeding full scale.
+Stock FireOS never writes this control **at all** (absent from
+`/system/etc/audio_device.xml` and from every `/system` binary), leaving the
+DAC at its 0dB reset default and taking user volume from AudioFlinger's
+software attenuation, which only ever attenuates — that is why native Alexa
+has no such distortion.
+
+Two things not to undo: `DEVICE_VOLUME_MAX`/`volumeMax` stay at 127 (both
+pinned by test), and the conversion lives in **one** place — `em_volume.py`,
+because `level / 175` was copy-pasted into `em_controller`, `em_esphome` and
+`em_api` with no test on any of them, which is how the wrong ceiling survived.
+The lost headroom **cannot** be bought back from `Ext_Amp_Gain` (ctl 13): that
+control is inert on this board — sweeping its full 6/12/18/24dB range moves
+the output 0.0dB while still reading its new value back, the same shape as the
+mute LED being on a different GPIO than Amazon's own HAL believed.
+`HP Driver Gain Volume` (ctl 62) *is* live (+18dB commanded → +18.1dB actual,
+THD 2.25%) if more output is ever wanted, but that is a taste call to make by
+ear, and the speaker's behaviour above stock level is unmeasured.
+
+The **physical buttons** traverse `volumeButtonFloor`(47, −40dB)…127 in 4dB
+steps rather than the whole control: the scale is dB-linear, so the bottom
+third is indistinguishable from silence and stepping across it spends presses
+to go nowhere. Silencing the device is the mute button's job. Explicit `Set()`
+calls are deliberately **not** floored — HA's volume 0.0 must still mean
+silent — and a press from below the floor lands *on* it, so one press always
+reaches audible.
+
 Volume is **state, not a setting** — it rides the config channel but has no dashboard control (the slider was removed 2026-07-25: `SeedVolume` ignores later pushes, so moving it did nothing until the device restarted and any real volume change overwrote it). It is listed in `em_config_sections.STATE_KEYS`, exempt from section scoping, and shown read-only on the Status tab.
 
 Volume persists through reboots **controller-side**: every device `volume_state` report is stored into the device's `startupVolume` config, and the device restores it via `Server.SeedVolume` on the **first config push per run only** (later pushes must not stomp live changes). Until seeded (or a local volume change makes the device authoritative), the device suppresses its connect-time `volume_state` report — reporting the boot-default level is what used to clobber the stored value on reboot. Mute is the opposite: **device-sovereign**, persisted locally in `/data/local/etc/echomuse/state.json` (survives OTA slot flips; written on toggle, restored at boot pre-connect — ADC mute immediately, red ring/button LED after LED init).
--- a/controller/Dockerfile
+++ b/controller/Dockerfile
@@ -78,6 +78,7 @@ COPY em_player.py .
 COPY em_config_sections.py .
 COPY em_recordings.py .
 COPY em_turnclock.py .
+COPY em_volume.py .
 COPY version.py .
 # ESPHome native API subpackage (frame protocol, message registry, vendored protobuf)
 COPY esphome/ ./esphome/
--- a/controller/device_payloads/start_server.sh
+++ b/controller/device_payloads/start_server.sh
@@ -119,6 +119,30 @@ sup_log() {
 
 sup_log "boot slot=$(readlink /data/local/bin/server 2>/dev/null)"
 
+# ── Native AFE opt-in (docs/native-afe-migration.md) ───────────────────────────
+# EM_NATIVE_AFE is read once by the Go binary at ITS OWN startup — a boot-time
+# choice, not a live config push (cmd/server.go's newAudioBackends: mic/speaker
+# are opened before any controller connection exists to push a config to).
+#
+# A marker FILE, not a hand-export in this script, is what makes the choice
+# durable: this script is re-synced against the canonical payload on every
+# OTA when it drifts (_sync_start_script) and a hand edit here would simply be
+# overwritten. The marker lives in DEVICE_TLS_DIR's directory, not this
+# script, so it survives exactly the syncs that would erase an edit here.
+#
+# Checked once, before the retry loop — `export` here is inherited by every
+# `/data/local/bin/server` launch for the rest of this script's run, so one
+# check covers every restart the retry loop performs. It does NOT take effect
+# on a running device until this script itself re-execs (a real reboot, or
+# `stop`/`start` of the echomuse service from a channel independent of the
+# server's own shell proxy — see the marker's controller-side helper for why
+# the shell proxy must not do this itself).
+NATIVE_AFE_MARKER=/data/local/etc/echomuse/native_afe.enabled
+if [ -f "$NATIVE_AFE_MARKER" ]; then
+    export EM_NATIVE_AFE=1
+    sup_log "native AFE opt-in marker present — EM_NATIVE_AFE=1"
+fi
+
 # ── Amp safety ────────────────────────────────────────────────────────────────
 # Mute + amp off whenever the server is not running. The server does this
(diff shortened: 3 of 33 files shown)


wilbowes:
Thanks — this is a serious piece of work, and the writeup made it reviewable.

Could you split two things out into their own PR? Both are live bugs on the current pipeline and I want them regardless of what happens with the AFE:

1. **Cap volume at the codec unity gain.** You offered this already. It looks worse than the PR describes: `volumeMax = 175` maps EchoMuse whole range straight onto ctl 61 raw, and `volumeStep = 17`, so the button steps land on 119, 136, 153, 170 — the top three all above unity. Against your measurements that is 65% THD at 153 and 89% at 170, reachable with three presses. Stock never writes the control, which explains why we sound worse than stock at high volume.

2. **The dashboard `Toggle` `disabled` prop.** Beamforming, Echo cancel and Auto gain rendering greyed, reading as off, and writing the opposite value on click is a bug on `main` today, independent of this branch.

The AEC and AGC findings are useful too, but they only bite on the AFE path, so they can stay here.

On the AFE itself — I am not ready to decide, for a reason your own PR raises. If ~65% THD was in the playback path, that is a strong candidate for the tin-can character that motivated this, and I would rather fix that first and re-listen than adopt a second pipeline to work around it. If it still sounds materially better through the AFE afterwards, that is a much stronger case and I will take it seriously.

Two other things I would rather be explicit about than leave implied:

**A second audio backend is a permanent support cost.** "Which pipeline are you on?" becomes the first question in every audio issue. That is not a reason not to do it, but it needs to be worth it.

**Your branch may answer an open question for us.** #117/#141 are unresolved: with a plug in the jack, mic capture stalls on a ~102.3s metronome and output degrades, and the live hypothesis is exactly that we take `pcm23p` and drive the codec ourselves, so accdet fires and the HAL that should respond is not there. If the AFE path makes that disappear, it confirms the hypothesis whether or not this merges. If you have a Dot and a headphone cable, that test is worth more to me right now than the merge decision. Note it can take up to ~15 minutes after insert to arm.

Happy to review the split PR quickly.

kdkavanagh:
I will get those PRs separated.

Re: quality /volume. The upper ranges of the volume scale were definitely clipping before this change, but audio _quality_ is improved across all ranges of volume, without requiring any EQ adjustment. I've personally verified this and am quite happy with the result. It's easy to toggle on/off - give it a try!

Re: two things to support. I agree, tho I also didn't want to PR a full scale replacement w/o the option to fallback, at least for now.  Longer term, it's really a question of the goals of this project: if it's to tinker with the audio pipeline, the existing custom pipeline works. But if it's to yield the best end-user experience, in my testing, it's hard to beat the logic that Amazon deployed with the device itself (which would also simplify this project by not needing a custom audio pipeline).

NicoWeio:
Quick naive question: Does this affect both the internal speaker and AUX?

kdkavanagh:
I haven't tested aux, but my intuition is yes, that would be handled by the original Amazon decoding pipeline

wilbowes:
Before the code review, there is one experiment worth running that could change what this PR *is*, and you are the only person set up to run it.

**With a plug in the headphone jack, does the ~102.3s mic-stall metronome still happen on the native path?**

Context, if you have not hit #117/#141: a plug in the jack degrades the whole audio subsystem for as long as it is present. Characterised on hardware — 27 measured capture gaps at **102.3s ±2s**, every outlier an exact 2× or 3×, each costing 0.3–2.3s of capture. Zero stalls in 51h on an unplugged device against 67 in three hours on a plugged one. Downstream it breaches `ping_timeout` and tears down the ESPHome satellite, which is what users actually report (#141: music pausing, restarting and skipping).

Everything programmable has been exonerated by measurement rather than argument: all 6016 tlv320aic32x4 registers identical between audible and silent, all 218 MediaTek SoC audio registers identical but for free-running counters, the full ALSA mixer identical across all three audible outcomes. It is also load-independent — 3-pole cable, 4-pole cable and headphones all fail. And it is **ours, not the hardware**: the same speaker and cable on a stock unlocked Dot still running Alexa works correctly.

The live hypothesis is the HAL. On stock, Amazon's audio HAL and mediaserver own the codec and coordinate with accdet on jack transitions. We `stop media`, take `pcm23p` and drive the codec ourselves — so accdet fires and the thing meant to respond is not there, and a retry that never completes looks exactly like a 102.3s metronome.

**This branch is the only change that puts the HAL back in the loop**, and your own comment says so explicitly:

> It does NOT run `stop media`: unlike PcmSpeaker, this backend needs mediaserver to keep owning the PCM, or the HAL's ASP front end never sees…

So the native path is close to a controlled test of that hypothesis, and it cuts both ways:

- **No metronome on the native path** — a months-old, independently-reported fault is explained and has a fix. That is a materially different PR from "better-sounding audio", and worth reviewing on different terms.
- **Metronome still present** — the HAL hypothesis is dead, which is nearly as valuable, because it is currently the only live theory and it would stop the next person spending an evening there.

Two things to note when running it: it arms some **minutes** after an insert — measured at 3 and at 14 — so wait the full window before reporting it absent, and it stops within one cycle of a removal. The tell is `[mic] clock` — healthy devices run a slightly negative deficit, a plugged one runs positive.

No rush on the review meanwhile. Separately: the volume cap is already on `main` as 2e344b6 and shipped in `controller-v2.20.0` / `v2.12.0`, with the THD table in the commit message — thank you for that, it closed #162. This branch will want a rebase to drop it.

There is also a direction question about taking a dependency on Amazon's HAL that is the maintainer's to answer rather than mine, and it is being discussed. Flagging it so the jack experiment is not read as the only thing standing between this and a merge.

kdkavanagh:
Headphone jack works great - Audio improvement seems to be even better there...

My take is that this native pipeline is a massive simplification of this app for little/no downside.

As an aside, and I mean this in a genuinely constructive way: You might want to consider taking a few minutes to humanize your PR/issue feedback.  As a professional software engineer, it's very obvious when I'm talking to claude - it's quite a tax to parse thru walls of claudeisms as a OSS contributor.

wilbowes:
That's a fair comment and something I have been mindful of. I am time poor when it comes to this and balancing building features, fixing problems, and addressing issues against "life" leaves me relying on Claude to review, analyse and respond to the majority of technical stuff. I'll look to do more of a BLUF/Smart Brevity approach going forward. Hopefully that will help, and apologies for the wall of text.

wilbowes:
**The reverse engineering here is solid work, and it's already paid off.** Two things you turned up are fixed on main: volume is capped at codec unity now (127, down from 175 — three presses used to reach 89% THD), and `Toggle` honours `disabled` in the handler.

The AFE itself I'm not going to merge. EchoMuse is committing to running without Amazon's stack so it can run on hardware that isn't an Echo, and adopting the HAL as the audio path takes us the other way. Having the PR in full is what let me decide that rather than leave it open another month. Sorry it took two rebases to arrive at a no.

What I'd rather do is take the target you were pointing at. `EQ_50.cfg` and `AFE.cfg` came off a device this week: stock boosts 150–250Hz by about 25dB and deletes everything below 115Hz the driver can't radiate. We shipped the deleting half in 2.20.2 and never built the boosting half, which is probably a good part of the tin-can character you were chasing. It's a filter stage we can write ourselves (#247), and your PR is why anyone went looking for the coefficients.

There is one thing I'd like from you. With a plug in the jack, our path stalls capture every ~102 seconds — 27 measured gaps, none in 51 hours unplugged — and it's the worst unfixed fault we have (#117/#141). You said the jack behaved well on the native path. Run an hour with a plug in and send me either a `grep 'capture stall'` on the log or a count of controller disconnects. That number tells us whether the HAL is what's saving it, and it's the one thing I can't measure from here.

`afe_probe` and both docs are worth having on their own — send them as a separate PR and I'll take them.

amitra93:
Commenting here because I have my own fork of EchoMuse with a few differences for now (full duplex streaming to a HACS addon, not using ESPHome, added stopwords, Sendspin, etc). I found this PR because I wasn't happy with how the ALSA path sounded. I adopted the changes in your fork @kdkavanagh, thanks for the work.

With this change, music sounds roughly similar to my Google Home Mini which I intend to replace.

wilbowes:
I think it's great that the AFE work has been adopted and is working well. I just made a very conscious and considered opinion to try and not use the Amazon parts in my work. I think the audio quality we have is greatly improved thanks to the non AFE fixes put in place. Really it's a learning journey for me with the premise that is Alexa can do it, why can't I on the same hardware - especially given I have root access. 

I'm an optimiser by nature too so I'll be looking to squeeze everything I can out of this device - and with Claude (and some clearly very clever community contributors) doing the heavy lifting, what could possibly go wrong! ☺️

amitra93:
Yep sounds good. I'll look at #247 fixes when they're in and switch back if it's equivalent.

PR #169, opened 2026-08-13
Never advertise a guessed address: detect SERVER_IP, or refuse

wilbowes:
A fresh Home Assistant add-on install with **Server IP** left blank advertised a checked-in literal — a developer's own machine — to the entire fleet over mDNS.

`config.yaml` ships `server_ip: ""`, `em_start.py` deliberately passes empty values through so "em_controller.py's own fallback applies", and that fallback was `10.10.1.236` in both `em_controller.py` and `em_esphome.py`.

### Why it produces no error anywhere

Nothing validates SERVER_IP against reality. The controller advertises whatever it is given, the device believes mDNS, and the result is a controller running perfectly that no device can reach. Measured on a live add-on before the fix:

```
mDNS  echomuse._emcontroller._tcp.local -> 10.10.1.236:8767   (nothing running)
add-on host 10.10.1.81                 -> 8767/8768/8770 all listening
```

The device provisions, appears for approval, and never arrives. Note that **appearing for approval proves only that the wizard ran** — `db.ensure_device_token` inserts the pending row itself (`approved=0`, `ip` and `firmware_ver` NULL) when the wizard pushes TLS credentials over USB, entirely pre-first-contact. So the one signal an operator would reasonably read as "it connected" is precisely the one that does not mean that. It reads as a broken device rather than an empty field.

Confirmed fixed in the field: setting the option to the add-on host's real address restored connectivity immediately.

### The fix

`em_hostip.resolve` decides, as a pure function:

| configured | detected | result |
|---|---|---|
| valid IPv4 | — | use it |
| empty/whitespace | available | use detected |
| empty | none | **refuse to start**, naming the fix |
| malformed | either | **refuse to start**, naming the value |

A malformed value is refused rather than repaired — `inet_aton` accepts a three-part form by reading the last part as a 16-bit tail, so a typo would otherwise be advertised as an address nobody typed. That is the same silent wrong answer this module exists to prevent.

Detection is a UDP `connect()` against **192.0.2.1 (RFC 5737 TEST-NET-1)**, which sets a route and reads back the source address without sending a packet. A public resolver address is the usual way to write this and works identically, but it is indistinguishable from a phone-home to anyone reading the file or a firewall log, and this project does not make outbound calls.

`test_no_literal_fallback_address_remains_in_the_module` greps the module for IP literals so the original bug cannot come back. It caught two of my own during development, including one in a comment.

### Also: the WebUSB error names its origin

Chrome's insecure-origin allowlist matches scheme, host **and port** exactly. Under the add-on the wizard page is served from Home Assistant's origin, not the controller's — so an operator who has already allowlisted the standalone dashboard gets no benefit and no way to tell why. The message now names `window.location.origin`, and under ingress no longer suggests `localhost:8768`, since `_ingress_only_middleware` rejects everything that is not the Supervisor gateway and that advice leads to a 403.

### Testing

371 controller tests pass, including a new `tests/test_hostip.py` (10 cases). The `test_deploy.py` COPY guard caught the missing Dockerfile line for the new module before it could crash-loop the container.


wilbowes:
Adds Home Assistant-native auth to the add-on, on top of the SERVER_IP fix already in this PR.

## Why

Under ingress, HA has already authenticated the person before the request reaches us. The EchoMuse login was a second lock on a door that was already locked — and the bootstrap token was worse: a one-time secret to go and read out of the add-on log before the dashboard would open at all.

Supervisor forwards the authenticated user as `X-Remote-User-Id`, and [its ingress proxy strips any client-supplied copies of those headers](https://github.com/home-assistant/supervisor/blob/main/supervisor/api/ingress.py) before proxying. So on a request that genuinely came from the gateway, they are proof of an authenticated HA session.

`POST /api/auth/ingress` mints a session from them; the landing page tries it before rendering any form. **The first HA user through the door becomes admin and never sees a token prompt.** The standalone container is untouched and keeps password + bootstrap, which is the right auth for the deployment that has no other.

## The security content, in one line

The header means something only when `INGRESS_ONLY` **and** the peer is Supervisor gateway. **Two conditions, never one** — on the standalone container that same header is attacker-supplied, so honouring it there would be an unauthenticated admin session on a dashboard that proxies a root shell to every device.

That decision is `em_ingressauth.decide`: pure and tested, for the same reason `em_linkauth.decide` is. It is pinned two ways in `test_deploy.py`:

- the call site must pass the live `INGRESS_ONLY` and `request.remote`, never literals
- nothing else in the tree may read `X-Remote-User-*`

## Decisions worth reviewing

**Keyed on the HA user id, never the name** (schema v18, `users.ha_user_id`, appended not edited). Names are editable in HA — keying on one hands a renamed user a fresh account, or hands somebody else account to whoever took the old name. These rows carry a password sentinel no bcrypt check can match.

**Only the FIRST HA user becomes admin; the rest are read-only until promoted.** `panel_admin` is documented as governing the sidebar *menu entry*, which is not the same claim as gating the ingress URL, and I could not confirm the stronger one from the docs. So the safe reading is assumed rather than the convenient one. It is now set explicitly in `config.yaml` so a change to Supervisor default cannot widen access silently. **If you can test with a non-admin HA account, that would settle whether this caution is needed** — if non-admins genuinely cannot reach ingress, auto-admin becomes defensible.

**No Sign out under ingress.** HA owns the session, so it would re-authenticate immediately and read as a broken button.

**HA display names contain spaces**, unlike every local username, and now reach the user table. `em_support` account redaction already covers them because it reads that table rather than matching a pattern — verified rather than assumed, with a test.

## Testing

395 controller tests pass (14 new in `test_ingressauth.py`, 3 new deploy-shape guards, 1 new support-redaction test). Both the JSX bundle and `index.html`s inline script were compiled and syntax-checked locally, since neither is exercised until the image build.

Not yet tested on hardware — worth a look before merge, particularly the first-run path on a fresh database.

PR #177, opened 2026-08-14
Cap volume at the codec's unity gain to stop the DAC clipping

kdkavanagh:
tinymix ctl 61 is the tlv320aic32x4 DAC *digital* volume: 176 steps of 0.5dB spanning -63.5..+24dB, with 0dB at index 127. volumeMax was 175 — the control's own maximum — so the top 27% of the range applied up to +24dB of digital gain to already near-full-scale PCM and saturated inside the DAC. That is the "audio is garbled above ~75% volume" report; 127/175 is 72.6%.

It is not a corner of the range you have to go looking for. volumeStep was 17, so from the boot midpoint the physical button lands on 119, 136, 153, 170 — the top THREE presses are all above unity, and 65% THD is three presses away on a device with no other way to turn it up. Every one of them was reported as "EchoMuse sounds worse than stock at high volume", which it is, for this reason.

Measured on hardware (1kHz at -6dBFS, recorded through the mic array):

  ctl 61   DAC out   fundamental   THD    h3 rel
     119     -10.0        -20.0    1.0%    -45dB
     127      -6.0        -16.1    1.5%    -42dB
     136      -1.5        -11.9    2.3%    -40dB
     153      +7.0         -8.0     65%   -3.7dB
     170     +15.5         -8.4     89%   -1.1dB
     170*    -10.0        -20.4    1.1%    -43dB   (*source scaled down)

Two independent signatures: the output level goes FLAT from 153 upward (136->153 gained 3.9dB where the mixer promised 8.5, 153->170 gained nothing) because it had stopped being able to get louder, and the odd harmonics explode while h2 barely moves — symmetric clipping, not speaker breakup. The last row is the control that rules out the amp and the speaker: same ctl 61 as the worst run, source scaled down to land at the reference acoustic level, and it is clean. So the gain stage is fine and it is purely source x gain exceeding full scale.

Stock FireOS never writes this control at all — it appears in no /system binary and not once in /system/etc/audio_device.xml — leaving the DAC at its 0dB reset default and taking user volume from AudioFlinger's software attenuation, which only ever attenuates. That is why native Alexa has no such distortion.

The lost headroom cannot be bought back from Ext_Amp_Gain (ctl 13): sweeping its full 6/12/18/24dB range moves the output 0.0dB while still reading its new value back, i.e. it is inert on this board. HP Driver Gain Volume (ctl 62) is live (+18dB commanded -> +18.1dB actual, THD 2.25%) but that is a taste call to make by ear, and the speaker above stock level is unmeasured.

Also:

- The physical buttons now traverse volumeButtonFloor(47, -40dB)..127 in 4dB steps instead of the whole control. The scale is dB-linear, so the bottom third is indistinguishable from silence and stepping across it spends presses to go nowhere; silencing the device is the mute button's job. Explicit Set() calls are NOT floored, so HA's volume 0.0 still means silent, and a press from below the floor lands on it.
- The LED arc spans the button band rather than 0..175, where the audible range crowded into the top LEDs.
- The conversion moves to em_volume.py. It was spelled `level / 175` in em_controller, em_esphome and em_api with no test on any of them, which is how the wrong ceiling survived; the ceiling now has one definition and is pinned by test on both sides of the wire.
- em_api._stored_volume validates with float() before calling, because em_volume returns 0.0 for bad input — right on the audio path, wrong where None means "not known".

Split out of #168 at the maintainer's request: this is a live bug on the current pipeline and is independent of the native-AFE work.

Claude-Session: https://claude.ai/code/session_01HqXJuZe2jCMpj4bRm8sV9z

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1263,6 +1263,44 @@ Reverting a section **discards** its stored values (`set_device_config_sections`
 
 ### Volume / mute persistence
 
+**The scale stops at the codec's unity gain, and that ceiling is load-bearing.**
+tinymix ctl 61 is the tlv320aic32x4 DAC *digital* volume: 176 steps of 0.5dB
+spanning −63.5…+24dB, with 0dB at index **127**. The firmware shipped
+`volumeMax = 175` — the control's own maximum — so the top 27% of the range
+applied up to +24dB of digital gain to already near-full-scale PCM and
+saturated inside the DAC. Measured on hardware 2026-08-13 (1kHz at −6dBFS,
+recorded through the mic array): THD 1.5% at index 127, 2.3% at 136, **65% at
+153, 89% at 170**, with the output level *flat* from 153 upward because it had
+stopped being able to get louder, and h3 at −1.1dB relative to the fundamental
+(very nearly a square wave). The control that isolates it: index 170 with the
+source scaled down to land at the same acoustic level reads 1.1% — clean — so
+the gain stage is fine and it is purely source × gain exceeding full scale.
+Stock FireOS never writes this control **at all** (absent from
+`/system/etc/audio_device.xml` and from every `/system` binary), leaving the
+DAC at its 0dB reset default and taking user volume from AudioFlinger's
+software attenuation, which only ever attenuates — that is why native Alexa
+has no such distortion.
+
+Two things not to undo: `DEVICE_VOLUME_MAX`/`volumeMax` stay at 127 (both
+pinned by test), and the conversion lives in **one** place — `em_volume.py`,
+because `level / 175` was copy-pasted into `em_controller`, `em_esphome` and
+`em_api` with no test on any of them, which is how the wrong ceiling survived.
+The lost headroom **cannot** be bought back from `Ext_Amp_Gain` (ctl 13): that
+control is inert on this board — sweeping its full 6/12/18/24dB range moves
+the output 0.0dB while still reading its new value back, the same shape as the
+mute LED being on a different GPIO than Amazon's own HAL believed.
+`HP Driver Gain Volume` (ctl 62) *is* live (+18dB commanded → +18.1dB actual,
+THD 2.25%) if more output is ever wanted, but that is a taste call to make by
+ear, and the speaker's behaviour above stock level is unmeasured.
+
+The **physical buttons** traverse `volumeButtonFloor`(47, −40dB)…127 in 4dB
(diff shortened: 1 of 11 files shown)


wilbowes (approved):
Reviewed against the code and the hardware claims. Approving — no correctness defects found, and the measurements are in the constant where they belong rather than in a commit message that will be unfindable in six months.

## Verified rather than taken on trust

- **`Set()` clamps to `volumeMax`**, so a stored `startupVolume` of 170 on an existing device lands at 127 on the next seed. No migration needed: stored values are raw ctl 61 indices whose meaning is unchanged, they are just capped. It self-heals, too — the device then reports 127 and the controller stores that.
- **Both mixed pairings are safe.** New controller + old firmware: HA 1.0 maps to 127, accepted. Old controller + new firmware: the controller sends up to 175 and `Set` clamps. Neither combination clips or errors, which matters because any pairing can occur in the field.
- **`showLEDs` at the boundaries.** Level 0 lights nothing (correct — silence is not one bar), below the floor gives 1, at the floor gives 1, at max gives 12.
- **The arithmetic in the comments.** `volumeStep = 8` is 4dB, and `(127−47)/8` is exactly the ten presses claimed.
- **No stragglers.** All three `/175` sites consolidated, `Dockerfile` gained the `COPY em_volume.py` the deploy-shape test requires, `CLAUDE.md` updated, and the `0–175` comments in `server.go`/`control.go` corrected rather than left to rot.

## Two consequences for the release notes, not for this PR

1. **Every user's Home Assistant volume percentage will jump.** The conversion moves from `level/175` to `level/127`, so a device physically at 100 was reported as 57% and now reports 79%. Nothing gets louder or quieter, but it looks like something changed on every device, and a volume-threshold automation will fire differently.
2. **Maximum volume genuinely drops.** Anyone running at 150–175 loses real loudness. They were at 65–89% THD so they were listening to a square wave, but *"it got quieter"* is what they will perceive first — and there is no headroom to offer them today, since this PR measures HP Driver Gain Volume (ctl 62) as live but explicitly leaves the speaker above stock level untested.

Both will be billed in the GA notes.

## One minor, pre-existing, not yours

`volumePct` at `dashboard.jsx:1629` is referenced as a fallback and defined nowhere — not in any Python, JSX or Go file. So the volume row falls through to `—` rather than showing the fallback it appears to offer. Unrelated to this change; noting it so it is not rediscovered as a regression.

Thanks for splitting this out of #168 and for the control run that isolates source × gain from the gain stage — that is the measurement that makes the conclusion stick.

PR #194, opened 2026-08-16
Serialise pipeline runs across a barge-in

wilbowes:
## The report

Barge-in "really isn't working" on Test Device 01. It isn't the detector — that fired correctly every time (0.932, 0.992, 0.997, 0.619 against a 0.50 threshold). The turn that follows the barge is what dies.

Five barge-ins in the EA log window, five `pipeline_refused`, zero audio captured:

```
09:07:59,836  Barge-in: starting interrupting turn
09:07:59,836  Starting ESPHome voice turn conversation_id=55f840aa…
09:07:59,840  Pipeline run ended
09:07:59,841  HA ended the run without starting a pipeline — releasing the turn
09:07:59,841  [TURN] trigger=barge-in outcome=pipeline_refused total=+4ms audio=0frames/0ms
```

## The cause

`VoiceAssistantEventResponse` carries `event_type` and a name/value `data` list — **no run identifier anywhere in the protocol**. It is designed for strictly one pipeline run at a time per connection, and the satellite is what enforces that.

HA doesn't enforce it for us. `handle_pipeline_start` clears the audio queue and cancels `_tts_streaming_task`, then assigns a new `_pipeline_task` *without cancelling the old one*. So a second start orphans the first run, which keeps emitting onto the same socket.

Barge-during-thinking cancelled locally and immediately started another turn, so the aborted run's tail landed on it: `RUN_END` 4ms in (read as terminal by the `_run_started` discriminator), a stale `STT_VAD_END` in the same millisecond, and the orphan's real ending 470ms later after we'd already closed the turn.

## The fix

**The abort exists and we weren't sending it.** `VoiceAssistantRequest(start=False)` → aioesphomeapi `handle_stop(True)` → HA `_abort_pipeline()`, which queues the audio sentinel *and* cancels `_pipeline_task`. `cancel_turn`'s docstring claimed otherwise, citing an `ESPHOME_SPEC.md §7.4` that no longer exists in the tree.

**HA acknowledges an abort with no wire message at all**, so the barrier has to be ordering, not timing: after an abort, discard every event until the next `RUN_START` — necessarily ours, since HA emits it only on pipeline creation, we created exactly one more, and the connection is processed in order.

Armed only by an abort and bounded to one turn, so HA's wake-word interception — a genuine `RUN_END` with no `RUN_START`, which `_ha_never_started` exists to catch and whose absence stalled the satellite setup dialog — is untouched.

The playback branch serialises through `abort_ha_run()`, which arms the barrier without marking the turn cancelled (the response *was* delivered). `RUN_END` follows `TTS_END` so it's usually consumed before the barge lands; "usually" costs the whole interrupting turn when it loses.

## Not covered

Barging in and then thinking better of it. Once aborted the original turn cannot be resumed — the protocol has no resume — so preserving it needs the abort deferred until speech is confirmed. Filed separately.

## Tests

13 new, each verified by reintroducing the bug it covers, including the two that would pass silently otherwise: tidying `_discard_until_run_start = False` into the reset block alongside its neighbours, and dropping `abort_ha=True` at the call site. 453 pass.

### DIFF
--- a/controller/Dockerfile
+++ b/controller/Dockerfile
@@ -80,6 +80,7 @@ COPY em_player.py .
 COPY em_config_sections.py .
 COPY em_recordings.py .
 COPY em_turnclock.py .
+COPY em_runbarrier.py .
 COPY em_volume.py .
 COPY version.py .
 # ESPHome native API subpackage (frame protocol, message registry, vendored protobuf)
--- a/controller/em_controller.py
+++ b/controller/em_controller.py
@@ -1066,11 +1066,20 @@ async def _barge_watcher(device: Device, playback_started: asyncio.Event):
                     device.cancel_event.set()
                     if in_playback:
                         await device.send_control({"type": "speaker_flush"})
+                        # HA's run is normally over by now (RUN_END follows
+                        # TTS_END, and we only reach playback at TTS_END), but
+                        # a barge in the first milliseconds of audio can beat
+                        # it. Serialise anyway — the interrupting turn is the
+                        # thing that pays if we lose that race.
+                        esphome.abort_ha_run(device.device_id)
                     else:
-                        # Nothing is playing — abort the in-flight HA
-                        # pipeline instead (local-only; a late HA result is
-                        # discarded on arrival).
-                        esphome.cancel_voice_turn(device.device_id)
+                        # Nothing is playing — HA is mid-pipeline, and an
+                        # interrupting turn is about to start on the same
+                        # connection. The protocol carries no run id, so the
+                        # old run MUST be aborted upstream first or its tail
+                        # events land on the new turn and kill it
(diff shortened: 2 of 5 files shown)


wilbowes:
Same rebuild as #196 — the tests imported `em_esphome` and CI caught it at collection.

The barrier is now `em_runbarrier.RunBarrier`. Worth the split beyond testability: both of its failure modes are silent — swallow too little and the interrupting turn dies, swallow too much and the satellite goes deaf.

One thing changed on the way. `end_turn()` looked like decoration, since `begin_turn()` also clears the flag — but a mutation showed the original test passed against a no-op `end_turn`. It is a real bound: events are dispatched whether or not a turn is in progress, and `begin_turn` would only clear a stuck barrier once another turn started, which is a turn whose own events are being discarded. Test rewritten to cover it.

All checks green.

Issue #186, opened 2026-08-16
Adding a device plays the setup jingle twice


wilbowes:
Adding a device in Home Assistant plays the setup jingle **twice**, on every device, every time. Home Assistant issues two connection tests a few milliseconds apart and we play both.

Cosmetic — both tests succeed and setup completes — but it is the first thing a new user hears from a device they have just provisioned, and it sounds like something stuttered.

## Evidence

Deterministic across every build tested on 2026-08-16 (`2.20.0-ea.2` through `ea.5`) and both devices. Home Assistant's log:

```
18:20:27.405 DEBUG [assist_satellite.connection_test] Request for connection test with id …
18:20:27.405 DEBUG [assist_satellite.connection_test] Request for connection test with id …
18:20:58.143 DEBUG [assist_satellite.connection_test] Request for connection test with id …
18:20:58.156 DEBUG [assist_satellite.connection_test] Request for connection test with id …
```

Two tests, **two different ids**. And ours, receiving them:

```
18:20:27,138 [esphome.80440C95] AnnounceRequest: media_id='…/ffmpeg_proxy/465b…'
18:20:27,152 [esphome.80440C95] AnnounceRequest: media_id='…/ffmpeg_proxy/465b…'
```

14ms apart. So we are playing exactly what we are told, twice — this is not our announce handling, and not a retry after a failure.

Both announcements sit on the **same** connection. HA's config flow probes with a bare `aioesphomeapi` client, disconnects, and reconnects as `Home Assistant 2026.8.x` — but that bounce completes *before* either announce, so it does not explain the pair either.

## Not to be worked around here

The obvious fix — suppress a second `AnnounceRequest` arriving within some window — is wrong. We cannot tell a duplicated connection test from a genuine second announcement, and swallowing a real one is a worse failure than playing a jingle twice: an announcement is something a user's automation asked for and expects to hear.

Any deduplication would also have to key on something. The two requests carry different connection-test ids, so id is no help; the `media_id` prefix matches, but that is an implementation detail of HA's ffmpeg proxy and not a contract.

## Likely upstream

This looks like Home Assistant firing its satellite connection test twice rather than anything in the ESPHome protocol. Worth reproducing against a real ESPHome voice satellite (a Voice PE) to confirm before reporting to `home-assistant/core` — if a Voice PE also jingles twice it is plainly theirs, and if it does not, something about how we present the satellite is provoking it and the investigation comes back here.

## Related

[...]

PR #196, opened 2026-08-17
Send AnnounceFinished when the announcement actually finishes

wilbowes:
Second of two from the HA protocol audit (#194 is the barge-in fix). Independent of it — branched from `main`, no shared files beyond `em_esphome.py` in non-overlapping regions.

## What was wrong

`VoiceAssistantAnnounceFinished` is HA's completion signal, and **HA blocks on it**:

- `assist_satellite/entity.py` → `async_internal_announce` sets `_is_announcing`, puts the entity in `RESPONDING`, and awaits `async_announce`, whose contract is *"should block until the announcement is done playing"*. A concurrent announce raises `SatelliteBusyError`.
- `esphome/assist_satellite.py` → `_do_announce` implements that by awaiting our reply through `send_voice_assistant_announcement_await_response`.

We replied synchronously in the message handler, before a byte had played. Consequences:

- `assist_satellite.announce` returns while audio is still playing.
- The entity drops out of `RESPONDING` immediately.
- Two chained announcements overlap on the device, because HA's `_is_announcing` guard has already been released.

## The justification didn't hold

The comment read *"state transitions are sent synchronously so the wizard doesn't time out"*. Checking it:

- `_ANNOUNCEMENT_TIMEOUT_SEC = 5 * 60`. Five minutes.
- The setup wizard's connection test doesn't wait on this message at all — `assist_satellite/websocket_api.py` fires its event when the device fetches the `CONNECTION_TEST_URL_BASE` media id.

So an honest reply cannot break the wizard.

To be explicit about scope: I checked whether this explains #186 (setup jingle playing twice) and **it doesn't** — that test is a single announce awaiting a URL fetch. #186 stays open on its own terms.

## Now

`AnnounceFinished` comes from `_run_announce`'s `finally`, on every path, and `success` reports whether the audio actually reached the speaker.

Not replying is strictly worse than replying `success=False` — it parks HA for five minutes holding `_is_announcing`, after which every announcement fails. `ANNOUNCE_TIMEOUT_S` (120s) makes us the side that gives up first. The layer below is already bounded (`audio_duration * 2 + 10s` in `_run_post_turn_playback`), which is what the layer below always looks like.

`MediaPlayerState.ANNOUNCING` still goes out synchronously — it describes the state we are entering, not one we have reached.

## Tests

9 new, `asyncio.run()` style to match the rest of the suite (pytest-asyncio isn't in the test environment). Verified by reintroducing the early reply — 7 of the 9 fail. Covers both directions: the reply must not come early, and it must always come (fetch failure, empty media_id, empty audio, no playback callback, wedged playback, closed transport). 449 pass.

### DIFF
--- a/controller/Dockerfile
+++ b/controller/Dockerfile
@@ -80,6 +80,7 @@ COPY em_player.py .
 COPY em_config_sections.py .
 COPY em_recordings.py .
 COPY em_turnclock.py .
+COPY em_announce.py .
 COPY em_volume.py .
 COPY version.py .
 # ESPHome native API subpackage (frame protocol, message registry, vendored protobuf)
--- a/controller/em_announce.py
+++ b/controller/em_announce.py
@@ -1,0 +1,96 @@
+"""
+em_announce.py — running an HA announcement to completion.
+
+`VoiceAssistantAnnounceFinished` is Home Assistant's completion signal, and HA
+BLOCKS on it. `assist_satellite.entity.async_internal_announce` documents
+`async_announce` as "should block until the announcement is done playing",
+holds `_is_announcing` and the RESPONDING state for its duration, and raises
+`SatelliteBusyError` if another announcement arrives meanwhile; the esphome
+integration implements that by awaiting the reply through
+`send_voice_assistant_announcement_await_response`.
+
+Two rules follow, and they pull in opposite directions:
+
+  * do not reply early — an early reply returns the service call while audio is
+    still playing, drops the entity out of RESPONDING, and lets two chained
+    announcements overlap on the device instead of queueing behind HA's guard;
+  * always reply — a reply that never arrives parks HA for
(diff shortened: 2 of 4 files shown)


wilbowes:
Rebuilt after CI: the tests imported `em_esphome`, which pulls in zeroconf, aiohttp and the database — the line the minimal controller test env deliberately draws.

Rather than widen that env, the sequencing moved to `em_announce.run`, the same split as `em_linkauth.decide` and `em_button.decide`. It earns it on its own terms: the two rules pull against each other (never reply early, always reply) and the code reads fine with either broken. `em_esphome` keeps the adapter — resolving the playback callback, and deciding whether the reply can still be written to the transport.

The one thing that could not move is pinned against the source: that the message handler does not answer the announce itself, which is the exact shape the bug took.

Verified by reintroducing both: the early reply fails 7 of 12, answering from the handler fails the shape test. All checks green.

PR #208, opened 2026-08-17
Add follow-up conversation window after voice turns

finik:
After any turn that produced a real spoken answer, play a short chime and reopen the mic for ~9s without requiring the wake word — treats a timely follow-up as a continuation (same conversation_id, so HA's agent keeps short-term context) and falls back to wake-word idle quietly if nothing is said.

- em_esphome.py: thread conversation_id and no_speech_timeout through run_esphome_voice_turn/trigger_voice_turn; trigger_voice_turn now returns (should_continue, no_speech_timeout_hit, conversation_id) instead of a bare bool.
- em_controller.py: synthesize a two-tone earcon, and extend the post-turn loop with a follow-up branch that reuses the existing continuation-turn mechanics.

### DIFF
--- a/controller/em_controller.py
+++ b/controller/em_controller.py
@@ -52,6 +52,7 @@ import collections
 import contextlib
 import json
 import logging
+import math
 import os
 import socket
 import struct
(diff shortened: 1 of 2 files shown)


wilbowes:
Thanks for this — you've clearly read the turn loop closely, and the `_run_voice_locked` state handling in particular is not a fun read. I've gone through it against both our code and Home Assistant's, and I'm going to decline it, but I want to show the working rather than just say no.

## Continuation already exists, and it's HA-driven

`INTENT_END` carries `continue_conversation`, we read it at `em_esphome.py:743`, and `em_controller.py:1684` re-triggers the turn with the mic restarted, stale frames drained and the ring re-armed. That's the `trigger_type = "continuation"` path — it's been in since v2.6.4 (July) and our turn history has 212 real continuation turns recorded against it.

So when HA's conversation agent decides to hold the floor, the mic already reopens without a wake word. The deliberate part is that HA decides. The agent knows whether it asked a question; we don't.

## The `conversation_id` change isn't a fix

This was the part I most expected to take, and it doesn't work — for two independent reasons.

First, on our side: `run_esphome_voice_turn`'s `finally` clears `self._conversation_id = ""` (`em_esphome.py:1142`) before `trigger_voice_turn` reads it. So the returned id is always `""`, which is falsy, and `conversation_id or str(uuid.uuid4())` mints a fresh one every turn exactly as before.

Second, and the reason it isn't worth repairing: **HA discards the field.** `ESPHomeAssistSatellite.handle_pipeline_start` takes `conversation_id` as a parameter (`esphome/assist_satellite.py:525`) and never forwards it — the `async_accept_pipeline_from_satellite` call at :601 doesn't pass it. Continuity is owned by the base entity instead: `AssistSatelliteEntity` keeps its own `_conversation_id`, opens a chat session against it on every run, reuses it while valid or mints a fresh one when it expires, and passes `session.conversation_id` into the pipeline (`assist_satellite/entity.py:503-526`).

So HA already maintains short-term context across a continuation chain, with its own session lifetime, and the id we send is ignored on arrival. There's no context being lost for us to restore.

## Why not the follow-up window itself

Separate from whether it works — it's a policy we don't want to own.

The ESPHome voice assistant protocol already carries a continuation mechanism and HA implements the decision behind it. An unconditional window is a second, competing mechanism that fires when the agent has explicitly not asked for one, and the two would need reconciling forever.

It also has a failure mode we've already observed. Continuation turns in our own history include a device answering a nearby television — the TV speaks into the reopened window, that produces a turn, which opens another. Under HA control that self-limits, because the agent stops asking. A window that opens after every answer doesn't: something is always talking, so the 9s timeout never arrives. Our data has this happening across three devices in different rooms.

And the mic sitting open for 9s after every answer is a posture change for a project whose default is that the mic opens on a wake word. With `saveUtterances` enabled it also writes those seconds to disk as WAVs. If it ever lands it needs to be a per-device config key defaulting to off, not fleet behaviour.

## Two things worth knowing if you reuse this elsewhere

- The follow-up branch keys on `not no_speech_timeout_hit`, but `_no_speech_timeout` is only set on the two genuine no-speech paths. `pipeline_refused`, `tts_error`, `no_tts` and `stream_timeout` all return `False`, so the chime plays and the mic opens for 9s after turns that said nothing at all. `trace.outcome == "ok"` is the condition the PR description actually describes.
- `_play_followup_chime` sleeps a fixed `SPEAKER_PRIME_SECONDS` after `stream_speaker` returns. A socket write completes near-instantly regardless of link speed, so that's a guess at playback end rather than a measurement — the device reports the real end via `playback_stats` (`device.playback_done`), which is what the LED ring and the `speaking` flag were both moved onto after making this same mistake.

Genuinely appreciate the effort here, and the `no_speech_timeout` plumbing is clean work. If continuation isn't firing for you in practice, that's a real thing worth chasing — ours stopped on 9 August and we haven't explained it yet. I'd rather find out why the existing mechanism went quiet than add a second one alongside it. Happy to compare notes on that in an issue if you're seeing it too.

PR #216, opened 2026-08-17
Fix TypeError in get_user_by_ha_id causing Ingress login to fail

### LINKED ISSUE #211: Ingress login fails with TypeError in em_db.py (_conn context manager)
### Description
When accessing EchoMuse via the Home Assistant sidebar (Ingress), the login fails with an HTTP 500 error during `POST /api/auth/ingress`. Instead of logging in automatically via Home Assistant, the UI falls back to the setup page / token setup.

### Steps to Reproduce
1. Install the EchoMuse Home Assistant add-on.
2. Click on the EchoMuse item in the Home Assistant sidebar.
3. Observe that authentication fails and the setup page is shown.

### Expected Behavior
Ingress authentication should automatically log the Home Assistant user in without requesting a setup token.

### Traceback / Log Output
```
2026-08-16 19:13:12,112 [ERROR] echomuse.api — Unhandled error in POST /api/auth/ingress
Traceback (most recent call last):
  File "/app/em_api.py", line 379, in _error_middleware
    return await handler(request)
           ^^^^^^^^^^^^^^^^^^^^^^
  File "/app/em_api.py", line 578, in _post_ingress_login
    token, role = await auth.login_via_ingress(identity)
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/em_auth.py", line 165, in login_via_ingress
    user = await loop.run_in_executor(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/concurrent/futures/thread.py", line 59, in run
    result = self.fn(*self.args, **self.kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/em_db.py", line 2093, in get_user_by_ha_id
    with _conn() as conn:
         ^^^^^^^
TypeError: function takes exactly 1 argument (0 given)
```

### Root Cause
In `controller/em_db.py` ([line 2093](https://github.com/wilbowes/EchoMuse/blob/273e679b9724bee4a699d5b1ed05d6b9ff7328a8/controller/em_db.py#L2093), inside `get_user_by_ha_id`), the code attempts to execute:
```python
with _conn() as conn:
```
1. _conn is the global sqlite3.Connection instance, not a callable function/context manager. Invoking _conn() causes Python to throw a TypeError.
2. Following the thread-safety design of em_db.py (where DB access is guarded by _db_lock), read queries should use the module's internal helper functions like _q1(sql, params) or lock access explicitly with with _db_lock:. ([line 795](https://github.com/wilbowes/EchoMuse/blob/273e679b9724bee4a699d5b1ed05d6b9ff7328a8/controller/em_db.py#L795))

lennart24:
### Description
This PR fixes a bug that caused the Home Assistant Ingress authentication to fail and fall back to the setup token page. 

When accessing the app via the Home Assistant sidebar, the `POST /api/auth/ingress` endpoint threw a `HTTP 500` error:
`TypeError: function takes exactly 1 argument (0 given)`

This should fix #211 

### Root Cause & Fix
In `controller/em_db.py`, inside `get_user_by_ha_id`, the global `sqlite3.Connection` object `_conn` was mistakenly called as a context manager (`with _conn() as conn:`). 

I fixed this by replacing the faulty block with the existing thread-safe `_q1` helper function, which  handles the `_db_lock` and the `fetchone()` logic.

### DIFF
--- a/controller/em_db.py
+++ b/controller/em_db.py
@@ -2090,10 +2090,8 @@ def create_user(username: str, password_hash: str, role: str = "readonly") -> in
 
 def get_user_by_ha_id(ha_user_id: str) -> Optional[sqlite3.Row]:
     """Return the user linked to a Home Assistant user id, or None."""
-    with _conn() as conn:
-        return conn.execute(
-            "SELECT * FROM users WHERE ha_user_id = ?", (ha_user_id,)
-        ).fetchone()
+    # Use the existing _q1 helper which safely handles _db_lock and returns a single row
+    return _q1("SELECT * FROM users WHERE ha_user_id = ?", (ha_user_id,))
 
 
 def create_ha_user(ha_user_id: str, username: str, role: str) -> int:


wilbowes:
Thanks — merging #218 instead, which is the same one-line change to `_q1` plus `tests/test_db_users.py` covering the lookup against a real migrated database. This function shipped broken and survived several releases precisely because nothing exercised it, so the test is the part that stops it happening again.

Your diagnosis in #211 is what found it, and the merge commit credits you for that. The write-up was exact — right down to naming `_q1` and the `_db_lock` bypass rather than just the `TypeError` — which is why this took minutes rather than an afternoon.

#217 is still open and I'll come back to it there.

PR #217, opened 2026-08-17
Fix ESPHome MAC collision for multiple devices by hashing DSN

### LINKED ISSUE #212: ESPHome MAC collision: _serialno_to_mac strips non-hex characters, resulting in identical MACs for different devices
### Description
When trying to add more than one Echo device to Home Assistant via EchoMuse, the ESPHome integration overwrites the existing device instead of creating a new one. The second device is also not automatically discovered. 

### Error Message from Home Assistant
Home Assistant clearly shows a MAC collision in the UI when adding the second device:
> "Device <device_name> with mac address <XX:XX:XX:XX:XX:XX> is already configured. The existing configuration will be updated."

### Root Cause
The bug is in `controller/em_esphome.py` inside the [`_serialno_to_mac(device_id)` function](https://github.com/wilbowes/EchoMuse/blob/c6f9addcc557c43ecec33e79f86c7777de82caf2/controller/em_esphome.py#L2557). 

The docstring assumes that `ro.serialno` is a purely hexadecimal string:
`# ro.serialno on the biscuit is a 12-char uppercase hex string`

However, Amazon Device Serial Numbers (DSNs) are alphanumeric. The function filters out any non-hex character:
```python
hex_chars = "".join(c for c in device_id if c in "0123456789ABCDEFabcdef")
```
If a user has two Echo devices from the same batch, their serial numbers often only differ in the last few characters, which are frequently non-hex letters (e.g., ending in ...FRK vs ...FRR).
Because K and R are stripped by the filter, the remaining string becomes completely identical for both devices. As a result, the function returns the exact same fake MAC address for both satellites. Since Home Assistant strictly uses the mac_address as the unique_id in its device registry, it assumes it's the exact same device and overwrites the existing entity.

### Proposed Fix
Instead of stripping non-hex characters (which destroys entropy), the code could hash the entire device_id to generate a stable, pseudo-random MAC address.

Example fix using MD5:
```python
import hashlib

def _serialno_to_mac(device_id: str) -> str:
    # Hash the unique alphanumeric serial number to get a predictable hex string
    hashed = hashlib.md5(device_id.encode('utf-8')).hexdigest()
    # Take the first 12 hex chars
    hex_chars = hashed[:12].upper()
    # Format as MAC address
    return ":".join(hex_chars[i:i+2] for i in range(0, 12, 2))
```

lennart24:
### Description
This PR fixes a bug (fix #212) where adding multiple Echo devices to Home Assistant via ESPHome overwrites the existing device instead of creating a new one. 

### Root Cause
The `_serialno_to_mac(device_id)` function stripped all non-hexadecimal characters from the alphanumeric Amazon Device Serial Number (DSN). Because devices from the same batch often only differ in non-hex characters at the end (e.g., ending in `...FRK` vs `...FRR`), the function generated the exact same MAC address for multiple satellites. 
Since Home Assistant strictly uses the `mac_address` as the `unique_id` in its device registry, it overwrote the existing entity instead of discovering a second one.

### The Fix
Instead of stripping characters (which destroys entropy), the function now hashes the entire `device_id` using MD5 to generate a stable, truly unique pseudo-MAC address for each device.

⚠️ **BREAKING CHANGE / NOTICE:** 
Because the algorithm for generating MAC addresses changes with this PR, already registered Echo devices in Home Assistant will change their MAC address. Users will need to re-add their devices to Home Assistant, as the old entities will become orphaned.

### DIFF
--- a/controller/em_esphome.py
+++ b/controller/em_esphome.py
@@ -2558,13 +2558,15 @@ def _serialno_to_mac(device_id: str) -> str:
     """
     Derive a stable MAC-format string from a device's ro.serialno.
 
-    ro.serialno on the biscuit is a 12-char uppercase hex string
-    (e.g. G0K0XXXXXXXX). We take the last 12 hex chars (or pad/truncate)
-    to build a MAC-style address for ESPHome's mac_address field.
-    This is cosmetic only — ESPHome's protocol uses it as a stable
-    device identifier in HA's device registry.
+    Amazon Device Serial Numbers (DSNs) are alphanumeric. Devices from the 
+    same batch often differ only in non-hexadecimal characters at the end 
+    (e.g., ...FRK vs ...FRR). Stripping non-hex characters destroys entropy 
+    and causes ESPHome MAC collisions in Home Assistant.
+    
+    Instead, we hash the entire device_id to generate a stable, pseudo-random 
+    MAC address to guarantee uniqueness per device.
     """
-    # Extract hex chars only, take last 12, pad with zeros if short
-    hex_chars = "".join(c for c in device_id if c in "0123456789ABCDEFabcdef")
-    hex_chars = hex_chars[-12:].upper().zfill(12)
+    import hashlib
+    hashed = hashlib.md5(device_id.encode('utf-8')).hexdigest()
+    hex_chars = hashed[:12].upper()
     return ":".join(hex_chars[i:i+2] for i in range(0, 12, 2))


wilbowes:
**Your diagnosis is right, and it's landed — you're credited as co-author on #254 (merged).** Apologies for the slow reply; that's on us, not on the PR.

**What you found is the real bug.** Stripping non-hex characters from an alphanumeric serial destroys exactly the entropy that distinguishes devices from one batch, and Home Assistant keys its device registry on the result. That's the whole fault, and hashing is the right answer.

**What #254 adds is storage, not a different fix.** Your change is correct but it moves *every* device's address at once — which is why you flagged that people would have to re-add their devices. Storing the value on the device row lets us keep the address HA already knows for devices that work today, and assign a new one only where there's an actual collision. Same fix, nobody re-adds anything.

**One thing we layered on:** a fixed `02:EC` prefix instead of the raw hash's first byte, so the address is always a valid locally-administered *unicast* MAC. A raw hash sets the multicast bit about half the time — worth saying the old code did that too, so it isn't something you introduced.

Leaving this open for you to close, since it's yours. Thanks for chasing it down — it was affecting anyone running more than one Echo, and it would have kept doing so.

PR #221, opened 2026-08-18
Build the controller image on PRs that change what goes into it

wilbowes:
## The gap

Nothing built the controller image on a pull request. `controller-build.yml` runs on `push: branches: [main]`, and every test job in `ci.yml` pins itself to `python-version: "3.12"` regardless of what the Dockerfile says. So a base-image or dependency bump arrived with **seven green checks and no evidence behind any of them**, and the first failure would have been on main, in the job that publishes.

Dependabot #127 is the worked example — `python:3.12-slim` → `3.14-slim`, green on all seven checks, and unbuildable. Measured on PyPI for the versions `requirements.txt` pins:

```
speexdsp-ns 0.1.2   cp37 … cp312     <- the binding constraint
onnxruntime 1.20.1  cp310 … cp313
numpy       2.4.4   cp311 … cp314
```

Neither builds from source in this image, so the bump does not degrade — it fails.

## The job

Narrow on purpose, since a full build is minutes:

- **Only when `Dockerfile` or `requirements.txt` changed.** Application code cannot break the build in a way the other jobs miss; a missing wheel is exactly what they cannot see.
- **amd64 only.** The break this catches is a missing wheel or an apt package that moved, both of which show on either architecture, and arm64 under QEMU costs far more than it adds. The release path still builds both.
- **No push, no registry login, no GHCR credentials.** It answers "does it build", nothing more.

Two shape decisions worth reviewing rather than assuming:

- **A job, not a separate workflow.** A workflow-level `paths:` filter would leave the check simply *absent* on most PRs, which is the one state a branch protection rule cannot distinguish from "not finished". As a job it always reports, including when it skips.
- **Plain `git diff`, not a paths-filter action.** This step decides whether a build runs, so it is not worth a third-party action holding repo credentials.

## dependabot.yml

Python major and minor bumps on the docker ecosystem are now ignored, with the wheel table written into the comment. Moving Python is a deliberate upgrade that starts by checking wheels, not a weekly PR — the same reasoning as the existing numpy and onnxruntime entries.

It also records why this ecosystem going quiet is **not** coverage: `python:3.12-slim` is a floating tag, so Debian security fixes arrive by rebuilding rather than by a PR here, and `security-scan.yml` is what watches for them.

## Verification

This PR touches `.github/` only, so the new job should report **skipped** on itself — which is the behaviour worth confirming in the checks below. #127 is the negative case and I have closed it against this evidence rather than merging it into a red build.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -44,12 +44,41 @@ updates:
   # ── Controller base image ───────────────────────────────────────────────
   # python:3.12-slim is where ffmpeg, curl and the whole Debian layer come
   # from — the majority of what any image scan reports.
+  #
+  # The Python version is NOT ours to choose: it is the highest one every
+  # pinned wheel publishes for, and today that is 3.12. Measured on PyPI
+  # 2026-08-18, for the versions requirements.txt actually pins:
+  #
+  #   speexdsp-ns 0.1.2  cp37 … cp312     <- the binding constraint
+  #   onnxruntime 1.20.1 cp310 … cp313
+  #   numpy       2.4.4  cp311 … cp314
+  #
+  # Neither builds from source in this image, so a bump past the lowest of
+  # those does not degrade — it fails to build. #127 proposed 3.14 and
+  # carried seven green checks, because until ci.yml gained the
+  # controller-image job nothing built the image on a PR at all.
+  #
+  # So majors and minors are ignored and moving Python is a deliberate
+  # upgrade: check the wheels first, then bump this and requirements.txt
+  # together.
+  #
+  # That leaves this ecosystem quiet, which is correct and worth stating so
+  # nobody reads the silence as coverage: `python:3.12-slim` is a FLOATING
+  # tag, so Debian security fixes arrive by rebuilding, not by a PR here.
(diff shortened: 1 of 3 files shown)


wilbowes:
**Both paths verified on this PR**, since a build trigger nobody has seen fire is worth nothing.

**Skip path** — the first push touched `.github/` only:

```
Dockerfile and requirements.txt unchanged
```

Job result **SUCCESS**, not absent. That is the behaviour the "job, not a separate workflow" decision exists for.

**Build path** — a temporary commit appended a comment line to `requirements.txt`:

```
controller/requirements.txt
#1 pulling image moby/buildkit:buildx-stable-1
#1 transferring dockerfile: 7.47kB done
Build the controller image = success
```

Real buildkit build, ~2.5 minutes cold with no cache to warm from. That commit is now reverted, so the net diff against main is the job and the config only:

```
 .github/dependabot.yml   | 29 ++++++++++++++++++++
 .github/workflows/ci.yml | 69 ++++++++++++++++++++++++++++++++++++++++++++++++
```

The revert leaves the two temporary commits in this branch's history; squash-merge collapses them.

PR #265, opened 2026-08-20
forge: run control, browser credentials, dashboard theming, and a published image

wilbowes:
Three things in the forge that previously needed a shell or a container restart.

### Stop a run

A job could only be waited out. It now runs in its own process group, and Stop sends SIGTERM to the group, then SIGKILL to whatever is left after 10s.

The process group is the point. `forge.py` is only the parent: `build` spawns openWakeWord's `train.py`, and *that* is what holds the GPU and writes the wake word's directory. Killing the parent alone leaves it running behind a UI reporting the run as stopped. Verified against a parent that ignores SIGTERM and spawns a child doing the same; both are reaped.

**Partial work is kept.** Clip generation is resumable and both later steps check what already exists, so the loop this enables is stop, adjust, train again, not start over.

### Change settings between runs

`PATCH /api/wakewords/{name}` edits `n_samples`, `steps` and `custom_negative_phrases`, exposed as *training settings* on each card.

The lines are rewritten in place rather than the YAML round-tripped: the template's comments explain every field, and a dump would silently delete them. The result is parsed before it is written, so a bad edit can't leave a config the trainer chokes on. `target_phrase` is deliberately not editable, because changing it invalidates every clip already generated, which is a new wake word rather than an edited one.

### Google credentials from the browser

Upload or paste the JSON key; no compose mapping to arrange. It's validated as a **service account** before being written, because an OAuth client secret is what people reach for first, looks close enough to pass a glance, and otherwise fails much later inside a job with nothing pointing back at the upload. Written 0600 through a temp file so it's never briefly world-readable, and only `project_id` and `client_email` ever travel back to the browser. Jobs read the path at spawn time, so a key uploaded after the container started works with no restart.

**Test connection** exists because local validation structurally can't see the usual failure: a valid key on a project where the Text-to-Speech API was never enabled.

### Theme

The UI now carries the dashboard's own token set, both themes, applied before first paint from the same `em-theme` key, so one pick covers both surfaces. Copied rather than reinvented so the two can't drift into two slightly different warm greys. Checked: no undefined `var()`, and both themes define an identical token set.

Also, the wake word list stops rebuilding itself over the top of what you're typing. It re-renders only when the markup changed, never while an input inside it has focus, and restores open panels afterwards.

### Worth knowing

The credentials endpoint writes a secret and there's no auth in front of it. Same trust model as the rest of this tool (it can already delete training data and run arbitrary jobs), but it's a secret at rest now, so the port should stay on the LAN. Noted in the README.

`oww_forge` isn't covered by CI, so none of this is tested by the suite. The endpoints were exercised by hand against a temp `FORGE_DATA` (validation rejects, permissions, no key leakage to the UI, config patch preserving comments and phrases, cancel reaping the grandchild).

### DIFF
--- a/.github/workflows/forge-release.yml
+++ b/.github/workflows/forge-release.yml
@@ -1,0 +1,239 @@
+name: Forge Release (Docker image)
+
+# Publishes the wake-word trainer as a built image, so training a custom wake
+# word does not start with a local build.
+#
+# The real reason is reproducibility rather than convenience. Every pin in
+# oww_forge/Dockerfile is load-bearing — piper-sample-generator v2.0.0 for its
+# flat layout, an openWakeWord SHA carrying a local patch, torch 2.7.1 — and a
+# local `docker compose build` re-resolves the floating layers underneath them
+# every time it runs. A published image is the only artifact that preserves
+# what was actually verified. Same argument as the controller's
+# promote-don't-rebuild flow, arrived at from the other direction.
+#
+# TWO VARIANTS, and the split is deliberate:
+#
+#   :X.Y.Z / :latest          CUDA, linux/amd64 ONLY
+#   :X.Y.Z-cpu / :latest-cpu  CPU torch, linux/amd64 + linux/arm64
+#
+# CUDA is amd64-only because there is no arm64 machine anyone would train on
+# with an NVIDIA GPU attached; building it would spend runner time producing
+# an image for hardware that does not exist.
+#
+# The CPU variant IS multi-arch, for Apple Silicon. Note what that does and
+# does not buy: Docker Desktop on macOS runs containers in a Linux VM with no
+# Metal/MPS passthrough, so training in a container on a Mac is CPU-bound
+# whatever we ship. arm64 does not unlock the GPU — it removes QEMU, which is
+# the difference between the documented overnight CPU run and something nobody
+# would sit through. Anyone wanting the M-series GPU has to run the trainer
+# outside Docker against torch's MPS backend, which is not what this ships.
+#
+# Tag prefix is `forge-v*`, kept clear of the plain `v*` firmware tags: OTA
+# release polling (em_api._fetch_latest_release) filters those for a `server`
+# asset, and the dashboard's controller notice prefix-matches `controller-v`.
+# Neither sees this one. No GitHub Release is created here, for the same
+# reason the controller creates none.
+
+on:
+  push:
+    tags:
+      - 'forge-v*'
+  # Builds WITHOUT publishing, so the workflow itself can be exercised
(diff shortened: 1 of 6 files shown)


wilbowes:
Second commit on this branch: **publish the forge as a built image.**

Training a custom wake word currently starts with a local build, and that build is also what undermines the pins — `piper-sample-generator` v2.0.0, an openWakeWord SHA carrying a local patch, torch 2.7.1 — because every `docker compose build` re-resolves the floating layers underneath them. A published image is the only artifact that preserves what was verified. Same argument as the controller's promote-don't-rebuild flow, reached from the other direction.

**Two variants.** `:X.Y.Z` / `:latest` is CUDA, **amd64 only** — there's no arm64 machine with an NVIDIA GPU anyone would train on. `:X.Y.Z-cpu` / `:latest-cpu` is CPU torch and **multi-arch**, which is the Apple Silicon answer.

Worth being precise about what arm64 actually buys, because it's easy to oversell: Docker Desktop on macOS runs containers in a Linux VM with **no Metal/MPS passthrough**, so training in a container on a Mac is CPU-bound whatever we ship. arm64 doesn't unlock the GPU — it removes QEMU, which is the difference between the documented overnight CPU run and something nobody would sit through. Using the M-series GPU means running outside Docker against torch's MPS backend, which isn't what this ships. I checked aarch64 wheels exist for torch and torchaudio 2.7.1 before committing to it, so the arm64 build compiles nothing but webrtcvad.

**The two CPU architectures build on their own native runners** (arm64 runners are free for a public repo) and push by digest, with the manifest assembled afterwards. Two jobs writing one tag would leave whichever lost silently absent from the manifest. The assembly step refuses to run on one digest — a manifest built from half the matrix publishes, pulls, and is simply the wrong architecture for half the people who asked, surfacing as `exec format error` on their machine rather than as a red build on ours.

**Runner disk is the obstacle worth knowing about.** A hosted runner has ~14GB free and CUDA torch doesn't fit. The reclaim step isn't tidiness; without it the build dies with a disk error that reads like a broken Dockerfile. Inline rather than a third-party action, so nothing outside this repo runs as part of a publish.

`workflow_dispatch` builds without publishing, so the workflow can be exercised without a tag — the expensive failure here is a tag that half-publishes.

Tag prefix `forge-v*` stays clear of the `v*` firmware tags OTA polling filters, and of the `controller-v` prefix the dashboard's update notice matches. No GitHub Release is created, same as the controller.

Also adds `.dockerignore`: `/data` is the training volume, mounted at run time, so without it a local build hands the daemon ~25GB of assets as build context — which reads as the build hanging before it runs an instruction.

**Untested by CI, as with the rest of this branch.** The workflow has never run; the first `workflow_dispatch` is the real check, and I'd do that before cutting `forge-v1.0.0`.

PR #277, opened 2026-08-22
Fix #276: firmware builds again under the pinned Go 1.24 compiler

### LINKED ISSUE #276: Device firmware cannot be built with its pinned compiler (go.mod needs 1.25, image has 1.24)
**The device firmware cannot be built with its own pinned compiler, and has not been able to since 2026-08-18.** `./compile.sh` fails, and the next `v*` tag would fail `release.yml` at the build step.

```
go: go.mod requires go >= 1.25.0 (running go 1.24.0; GOTOOLCHAIN=local)
```

## Cause

`device/go.mod` says `go 1.25.0`. The pinned compiler image carries **Go 1.24.0** and has `GOTOOLCHAIN=local` baked into its environment, so it will not download a newer toolchain — deliberately, since that pin is what stops a third party's floating tag changing the compiler under us.

The bump arrived in **#128, a Dependabot `golang.org/x/sys` bump merged 2026-08-18**, which raised the `go` directive as a side effect. Nothing in CI catches it: the host jobs run `go test`/`go vet` on amd64 with GitHub's own toolchain, and `device/CLAUDE.md` already says so —

> `go vet` and the host tests prove nothing here — they run on amd64 with the host toolchain, and this image is only ever exercised by compile.sh and release.yml.

## Why nobody noticed

The last firmware release was **v2.12.0 on 2026-08-17** — the day before the bump. No `v*` tag has been pushed since, so the only thing that exercises the compiler has not run.

## Options

1. **Lower `go.mod` back to `go 1.24.0`.** Reversible, needs no hardware, and is arguably just undoing a side effect nobody asked for. Worth checking whether anything in the tree actually needs 1.25 language or stdlib features first — nothing obvious does.
2. **Move the compiler pin to an image with Go 1.25.** This is the one that needs care: `device/CLAUDE.md` says moving the pin requires *a real device in the loop*, because the failure mode is a binary the hardware refuses to run and that is the one thing not recoverable from the dashboard.
3. Set `GOTOOLCHAIN` to allow a download in the build. **Not recommended** — it silently reintroduces exactly the "compiler changes under us with no PR" problem the digest pin exists to prevent.

Option 1 unless something needs 1.25.

## Guard worth adding either way

Nothing fails until someone pushes a release tag, which is the worst moment to discover it. A CI job that builds the firmware in the pinned compiler image on every PR would have caught this on #128 — that is what the image is for, and it is currently exercised only by a path that runs a handful of times a month.

## Impact right now

Blocks 3.0.0 (#272): the output chain landing in `device/` compiles and passes tests, but cannot be built into a shippable binary with the pinned toolchain. It was verified with a `GOTOOLCHAIN=go1.25.0` override, which is a compile check and **not** a shippable artifact.

chr-braun:
The firmware can be built with its pinned compiler again, and a change that breaks it now goes red in CI instead of failing at release time.

- Lowering only the `go` directive is not enough: x/sys v0.42.0+ declare `go 1.25.0` themselves, so both move back together (`go 1.24.0`, x/sys v0.41.0 — verified against the module proxy as the newest compatible version).
- Nothing in the tree needs 1.25 language or stdlib features.
- New `device-firmware-build` CI job builds the compiler image from its pinned Dockerfile and compiles `./cmd/` inside it, using release.yml's invocation. It runs unconditionally: between merge and release tag there is no other check that touches the pin.

Fixes #276

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -245,6 +245,57 @@ jobs:
           cache-from: type=gha
           cache-to: type=gha,mode=max
 
+  device-firmware-build:
+    name: Device firmware builds in pinned compiler
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v7
+        with:
+          submodules: true  # GoTinyAlsa replace directive in device/go.mod
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v4
+      # Why this job exists: #276. Dependabot #128 raised device/go.mod to
+      # Go 1.25 while the pinned compiler image carries 1.24, and NOTHING
+      # noticed for weeks -- every other device job (tests, vet, vulncheck)
+      # runs on the host toolchain, which resolves a newer Go silently.
+      # The pinned image is exercised by compile.sh (local, manual) and
+      # release.yml (tags only, a handful of times a month), so the first
+      # symptom of a broken pin was a release tag failing at the build step.
+      # A module graph that needs a newer Go than the image ships is caught
+      # here, on the PR that causes it, minutes rather than at tag time.
+      #
+      # No path filter, unlike the controller-image job below: that one is
+      # covered on main by controller-build.yml, but the firmware has NO
+      # build between "merged to main" and "release tag", so this job is
+      # also the only continuous check on direct pushes to main. It runs
+      # unconditionally; the GHA cache keeps the compiler image layers warm,
+      # so the steady-state cost is the go build itself.
+      - name: Build the pinned compiler image
+        uses: docker/build-push-action@v7
+        with:
+          context: device/compiler
+          push: false
+          load: true
+          tags: echomuse-compiler:ci
+          cache-from: type=gha
+          cache-to: type=gha,mode=max
+      # Same invocation as release.yml, minus the version ldflags: a CI run
+      # has no v* ref_name to embed, and BuildUnix is only meaningful for a
+      # binary a device will boot. Everything else must stay identical --
+      # this job exists to fail exactly where the release would.
+      - name: Compile the firmware
+        run: |
+          mkdir -p device/build
+          docker run --rm \
+            --entrypoint bash \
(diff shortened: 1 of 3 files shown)


wilbowes:
Closing this one as already fixed — but the pin correction in it was right, and it is in `main` now.

Eight of the nine you sent overnight are merged. Thank you for them, and for writing them to fit the codebase rather than around it: the issue references, the tests, and the reasoning in the commit messages made the review much faster than the line count suggests.

On this one specifically: the same fix landed the previous morning in `ab31aef`, on a branch, which is why you had no way of seeing it. That commit pinned `golang.org/x/sys` to **v0.40.0** and asserted in its own message that v0.40.0 was the newest release still declaring `go 1.24.0`. It is not. Checked against the module proxy:

```
v0.40.0 -> go 1.24.0
v0.41.0 -> go 1.24.0
v0.42.0 -> go 1.25.0
```

You pinned v0.41.0, which is correct, and `main` now carries it with a commit crediting where the catch came from. Dependabot has since been told to leave that module alone on minor bumps, so the same PR does not have to be written a third time — the note there says to lift it when the compiler pin moves, which needs hardware.

The CI job kept was the branch's, for one reason worth naming since your version is otherwise the same idea: it also asserts the output is an ARM binary. A build that silently produced a host binary would pass the build step and ship something the hardware refuses to run, which is the failure that is not recoverable from the dashboard.

Two other things came out of the batch and are worth flagging back to you:

- **#282** — the `barged` outcome is right and the reasoning about #250 hiding in the stats was correct. The two new assignments dropped the `if trace:` guard that every sibling in that function carries, on a parameter that defaults to `None`. Not reachable today, since the only caller always constructs a trace, so it was restored in a follow-up commit rather than sent back.
- **#278 and #283** both insert constants immediately after `SOURCE_STALL_MS`, so they conflicted with each other on merge. Trivially resolved, just worth knowing if you stack player changes again.

If you want somewhere to land next, the backlog is now labelled — `ready` is the set with nothing external blocking it, and `needs-hardware` is the set that cannot be settled without a device in the room.

PR #278, opened 2026-08-22
Fix #258: verify certificates on https media URLs, and let a refused decoder speak

### LINKED ISSUE #258: Media URLs are fetched over HTTPS without verifying the certificate
ffmpeg in the controller image defaults `-tls_verify` to **0**, so every media URL fetched over HTTPS is accepted without checking the peer certificate.

```
$ ffmpeg -h protocol=tls
  -tls_verify  <int>  ED......... Verify the peer certificate (from 0 to 1) (default 0)
```

Confirmed by measurement: an HTTPS server with a certificate from an untrusted private CA is fetched successfully by ffmpeg, and rejected by Python on the same URL in the same container.

`em_player` passes URLs straight to ffmpeg (`-i <url>`), so this covers everything the media player plays — Music Assistant streams, radio, anything `play_media` is handed.

## How much this matters

Less than it sounds, and more than nothing. Most of these URLs are LAN-local, and the content is music rather than credentials. But `play_media` will fetch whatever URL it's given, including internet radio and remote streams, and "we accept any certificate" is not a property anyone would choose deliberately. It's currently true by accident of an ffmpeg build default.

Found while adding `EM_EXTRA_CA_CERT` (#258) — it's the reason media playback was *not* among the symptoms the reporter saw, which was initially confusing.

## Shape of the fix

Add `-tls_verify 1` to the ffmpeg invocations in `em_player`.

**The reason this isn't a one-line change:** it can only break setups that work today. Anyone serving media over HTTPS with a self-signed or private-CA certificate is currently fine and would stop being fine. That's now recoverable — `EM_EXTRA_CA_CERT` installs into the system store, which GnuTLS-linked ffmpeg reads (verified) — but it needs to land *after* people have that option, and the failure mode wants a clear log line rather than a bare ffmpeg error.

Worth checking at the same time whether the TTS fetch path should pin anything, and whether a per-device or global opt-out is warranted for someone with a stream they can't fix.

chr-braun:
Media URLs fetched over HTTPS are now certificate-verified, and an ffmpeg that refuses a stream says so in the log instead of dying silently.

- `-tls_verify 1`, input-scoped like `-ss` and only for https — other protocols would silently ignore it.
- Lands after #259 deliberately: verification can only break setups that were working by accident, and those recover via `EM_EXTRA_CA_CERT`.
- The decoder's stderr is now captured; any abnormal exit logs its last lines, and a certificate failure points at `EM_EXTRA_CA_CERT`. docs/configuration.md notes that media verifies too.

Fixes #258

### DIFF
--- a/controller/em_player.py
+++ b/controller/em_player.py
@@ -45,6 +45,7 @@ from __future__ import annotations
 
 import asyncio
 import logging
+from collections import deque
 
 import em_eq
 import em_limiter
@@ -136,6 +137,15 @@ SEEK_STALL_S    = 5.0
 # 500ms is well inside the ~4s lead, so a stall this size is not yet audible —
 # which is the point of catching it here.
 SOURCE_STALL_MS = 500.0
+
+# How many ffmpeg stderr lines to keep for the failure log. -loglevel error
+# means anything that IS on stderr is meaningful; five lines covers every
+# ffmpeg diagnostic this player has needed to explain a dead decoder.
+STDERR_TAIL_LINES = 5
+# On an abnormal decoder exit, how long the finally block waits for the
+# stderr drain to deliver the last lines before logging without them. Only
+# failure paths pay this; pause/stop and natural end cancel immediately.
+DRAIN_GRACE_S = 0.5
 
 
 IDLE, PLAYING, PAUSED = "idle", "playing", "paused"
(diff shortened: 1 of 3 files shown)


wilbowes:
Merged as `fe59504` — thanks.

PR #279, opened 2026-08-22
Fix #159: update_check_interval=0 means OFF, not a busy loop

### LINKED ISSUE #159: update_check_interval: 0 spins the release poll loop instead of disabling it
`update_check_interval` is the obvious way to stop the controller contacting GitHub, and setting it to `0` does the opposite of stopping it.

```python
interval = int(db.get_config("update_check_interval", "3600") or 3600)
await asyncio.sleep(interval)
```

`db.get_config` returns a **string**, so the `or 3600` fallback only fires on an empty value. A stored `"0"` is a truthy string, so it survives to `int("0")` → `0` → `await asyncio.sleep(0)`, and the loop spins as fast as the event loop allows — hammering `api.github.com` until it rate-limits, and burning CPU that the loop-lag instrumentation exists to protect.

Verified:

```
stored "0"  -> interval 0      (busy loop)
stored ""   -> interval 3600   (fallback works)
```

It affects both poll loops, since `_fetch_latest_release` and `_fetch_controller_release` share the same sleep.

## Why it matters more now

#158 documents that the hourly GitHub release check is the controller's only outbound connection. The reader most likely to act on that is the reader who wants it off — and `0` is the first thing anyone tries. Today that turns one request an hour into a flood, which is the worst available outcome for exactly the person who was trying to be careful.

## Suggested fix

Treat `0` (and any non-positive value) as **disabled**: skip the fetch and sleep a fixed interval before re-reading the config, so re-enabling it from the dashboard still takes effect without a restart. That makes the documented knob do the obvious thing, and it is what the privacy section should be able to point at.

Worth guarding with a test — the failure is silent, and its symptom (GitHub rate-limiting) shows up somewhere else entirely.

chr-braun:
Setting `update_check_interval` to `0` now stops the GitHub poll instead of turning it into a busy loop against api.github.com.

- `db.get_config` returns a string, so `"0"` was truthy and survived the `or 3600` fallback into `asyncio.sleep(0)` — the worst outcome for exactly the reader who set it to stop the controller's only outbound connection (#158).
- Both call sites read through one parser: `<= 0` means disabled (poll parks at 60s so re-enabling needs no restart), non-numeric values fall back to 3600 instead of killing the poll task, positive intervals get a 1s floor.
- Tests run the shipped function sources in a stub namespace, since em_api is deliberately not importable in the minimal test env.

Fixes #159

### DIFF
--- a/controller/em_api.py
+++ b/controller/em_api.py
@@ -3162,6 +3162,26 @@ async def _push_log_event(
 
 # ─── GitHub release fetching ──────────────────────────────────────────────────
 
+def _update_check_interval() -> int:
+    """
+    The parsed update_check_interval; <= 0 means DISABLED (#159).
+
+    db.get_config returns a STRING, so a stored "0" is truthy and survived
+    the old `or 3600` fallback into asyncio.sleep(0) — turning the obvious
+    way to stop the controller contacting GitHub (#158 documents this poll
+    as its only outbound connection) into a busy loop against api.github.com
+    until it rate-limits. The worst available outcome for exactly the reader
+    who set it carefully. A non-numeric value used to raise out of the poll
+    loop and kill the task outright; it now falls back to the default with
+    a line in the log.
+    """
+    raw = (db.get_config("update_check_interval", "3600") or "3600").strip()
+    try:
+        return int(raw)
+    except ValueError:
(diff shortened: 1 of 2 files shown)


wilbowes:
Merged as `832bd22` — thanks.

PR #280, opened 2026-08-22
Fix #255: one bad message no longer costs the device its connection

### LINKED ISSUE #255: One bad control message tears down the whole device connection
An exception anywhere in the control-plane dispatch chain escapes to the handler's `except Exception`, which logs one line and lets the connection close. The device then reconnects, and everything downstream churns with it — ESPHome satellite server down and up, BLE proxy down and up, data plane closed, HA's connection dropped.

Observed on 2026-08-20 (fixed cause in #255): a missing attribute while handling a **`playback_stats`** message — a pure telemetry report — cost the device its connection and reconnected it on every barge-in.

```
[control] Handler error: 'Device' object has no attribute 'playback_send_ms'
[control] Device disconnected: G090LF1180440C95
[esphome.80440C95] Server stopped
[bleproxy.80440C95] Server stopped
[data] Data connection closed
```

## Why this is disproportionate

The messages arriving on this plane are not equally important. A `pong`, an `ambient_light` reading and a `playback_stats` report are all telemetry: dropping one costs a data point. `register` and `mute_state` are not. Treating every one of them as fatal to the connection means **the least important message in the protocol can disconnect the device**, and the blast radius is the entire per-device stack.

It also lengthens every incident. The visible symptom is a reconnect loop, which points the investigation at the network — the actual fault was one attribute in one handler.

## Shape of the fix

Wrap the per-message dispatch so an exception logs with a traceback and continues, rather than propagating. Two things worth getting right:

- **Log the message type and the traceback.** The current line gives neither, so a recurrence is as opaque as this one was; it took a stack trace to find.
- **Do not swallow `CancelledError`**, and consider whether a `register` failure genuinely should still be fatal — arriving in an unknown state is different from failing to record a statistic.

Not done at the time because the dispatch chain is ~370 lines and re-indenting it inside a `try` is a mechanical change that wants doing carefully, not during an incident.

Note the same shape exists on the data plane handler (`[data] Handler error`), which is worth reviewing at the same time.

chr-braun:
One bad control-plane message no longer costs the device its connection.

A missing attribute while handling a `playback_stats` telemetry report took the ESPHome satellite, the BLE proxy and the data plane down with it and reconnected on every barge-in (observed 2026-08-20). The dispatch now catches per message, logs type + traceback, and moves on; CancelledError still propagates, and register stays fatal deliberately. The data plane gets the same guard with a 5s rate limit — it runs at frame rate, so a persistent fault must not flood the log at 23 tracebacks/second.

Fixes #255

### DIFF
--- a/controller/em_controller.py
+++ b/controller/em_controller.py
@@ -2830,381 +2830,401 @@ async def handle_control(ws: WebSocketServerProtocol, secure: bool = False):
 
                 msg_type = msg.get("type")
 
-                if msg_type == "button":
-                    await handle_button_event(device, msg)
+                # One bad message must not cost the connection (#255):
+                # the LEAST important message in the protocol used to be
+                # able to disconnect the device - a missing attribute while
+                # handling a playback_stats telemetry report took the
+                # ESPHome satellite, the BLE proxy and the data plane down
+                # with it, and reconnected on every barge-in. A handler
+                # failure now logs message type + traceback and moves on.
+                # Deliberate exceptions:
+                #   - CancelledError propagates: teardown is not a message
+                #     fault.
+                #   - register runs BEFORE this loop and stays fatal -
+                #     arriving into an unknown state is different from
+                #     failing to record a statistic.
+                try:
+                    if msg_type == "button":
+                        await handle_button_event(device, msg)
 
-                elif msg_type == "ambient_light":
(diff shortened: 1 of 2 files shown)


wilbowes:
Merged as `012a43f` — thanks.

Issue #243, opened 2026-08-19
Design: should the output chain run on the device instead of the controller?


wilbowes:
**The output chain (EQ, bass guard, limiter) runs on the controller, so every setting change is heard about four seconds late, and the chain cannot respond to anything that happens after the audio has left.**

Measured 2026-08-19 while tuning the bass guard: an EQ change is audible ~4s after it is pushed. That is `em_player.LEAD_S` (4.0s) — the feed is paced that far ahead of realtime so a link stall costs margin instead of a gap, and the device holds ~5.5s of buffer behind it. Processing happens when the audio is generated, so the listener hears it a lead-time later.

## Why this is the same problem ducking already solved

Ducking had to be device-side for exactly this reason, and CLAUDE.md states it plainly:

> `LEAD_S` = 4.0s means the next four seconds of music are already in the device's buffer when a wake word fires, so **audio that has left the controller cannot be ducked by the controller**.

The output chain is in the same position. It is also, conceptually, *speaker correction* rather than a content effect — the bass guard exists because of the driver's excursion limit and the limiter because of what the DAC does above full scale. Corrections for a physical device arguably belong as close to that device as possible, and would then apply to everything it plays, including the mixed music+voice output.

## What it would cost

- **Not CPU.** Eight biquads, an LR4 crossover and a look-ahead limiter at 48kHz mono is a couple of million operations a second — low single-digit percent of one core against the ~18-20% the mic pipeline already uses. That is an estimate and should be measured on hardware before anyone relies on it, but it is not close enough to be the deciding factor.
- **Two implementations of the same DSP.** `em_eq`, `em_mbc` and `em_limiter` would need Go counterparts kept bit-compatible with the Python ones, which are unit-tested and measured. Duplicated logic is this project's most reliable source of bugs — two config.yaml files (#160), the dashboard's `CONFIG_SECTIONS` mirror, and the two config mirror sites that caused #241.
- **It is all-or-nothing.** The chain order is load-bearing (EQ → guard → limiter; limiting first spends gain reduction on bass about to be discarded), so the cheap EQ cannot move on its own.
- **Both paths would ship.** Firmware without the capability still needs the controller-side chain, per the degrade-to-old-behaviour rule — so this adds a path rather than replacing one, at least until the fleet has moved.

## What it would buy

- Setting changes audible immediately, which matters while tuning and not otherwise.
- Correction applied to everything the device plays, not only what the controller feeds it.
- One less thing the controller does per stream.

[...]

PR #281, opened 2026-08-22
Fix #275: bypassed clipping gets its own counter

### LINKED ISSUE #275: em_limiter's clipped counter reads as a bug when the limiter is simply off
`em_limiter`'s `clipped` counter is documented as meaning the limiter has a bug, and it goes non-zero in normal operation whenever the limiter is disabled.

The module docstring says:

> **A final clip that must never engage.** Kept as a backstop, and it is counted: if `clipped` is ever non-zero the limiter has a bug, and a silent backstop is how the original problem survived.

That reading is correct while the limiter is **limiting**. It is wrong while it is **bypassed** — where the limiter is supposed to do nothing, so a boosted EQ ahead of it legitimately pushes samples past the ceiling for the final `np.clip` to catch. `_emit` counts on both paths (`em_limiter.py:253`), so the counter cannot distinguish the two.

## Measured

Found while porting the chain to Go (#243). Two fixture cases run hot material with `limiterEnabled` false for part of the sequence:

| case | clipped samples |
|---|---|
| `switch_limiter_on` | 7 |
| `sweep_params` | 86 |

Both reproduce bit-for-bit in the Go port, so this is the reference's behaviour and not a porting artifact. Nothing is audibly wrong: the samples are clipped by the backstop exactly as intended for a signal nobody is limiting.

## Why it is worth fixing

The counter exists so a silent backstop cannot hide a real fault, and the instruction attached to it is "non-zero means a bug". Anyone acting on that while `limiterEnabled` is off investigates a limiter that is behaving correctly — and `limiterEnabled` is a supported setting, even though it defaults to true.

The same shape as the `nsAsr`/`0.00dB` distinction already documented in `controller/CLAUDE.md`: an instrument whose reading is identical in two states that want opposite investigations.

## Options

- **Count only while enabled.** Cheapest, and makes the docstring true as written. Loses the ability to see that the EQ is clipping with the limiter off, which is arguably worth knowing on its own.
- **Two counters** — `clipped` (while limiting) and something like `clipped_bypassed`. Keeps both signals, costs a field.
- **Leave the behaviour, fix the docstring.** Honest and free; the counter stays ambiguous.

No strong preference. Whichever is chosen should also land in `device/internal/outchain/limiter.go`, which currently keeps the count deliberately faithful so the two implementations stay comparable, with the caveat documented and pinned by `TestBypassedLimiterCanLegitimatelyClip`.

chr-braun:
The limiter's clip counter can no longer read as a bug while the limiter is simply off.

`clipped` is documented as "non-zero means a bug" — true while limiting, wrong while bypassed, where a boosted EQ legitimately pushes samples past the ceiling (#275 measured 86 across one fixture). Two counters now: `clipped` counts only while limiting, `clipped_bypassed` keeps the EQ-running-hot signal without the false alarm.

Note for #243's Go port: `device/internal/outchain/limiter.go` is not on main yet; when it lands it should follow this split rather than the faithful single counter.

Fixes #275

### DIFF
--- a/controller/em_limiter.py
+++ b/controller/em_limiter.py
@@ -46,8 +46,13 @@ per-sample Python loop is not available to us.
    across a long stream.
 
 3. **A final clip that must never engage.** Kept as a backstop, and it is
-   counted: if `clipped` is ever non-zero the limiter has a bug, and a
-   silent backstop is how the original problem survived.
+   counted: if `clipped` is ever non-zero while the limiter is LIMITING,
+   the limiter has a bug, and a silent backstop is how the original
+   problem survived. While the limiter is BYPASSED the same backstop
+   legitimately catches a boosted EQ's output — that goes to
+   `clipped_bypassed`, so the two readings cannot be confused (#275):
+   identical counts in both states would send someone hunting a bug in a
+   limiter that is behaving correctly.
 
 STATE
 -----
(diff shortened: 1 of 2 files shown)


wilbowes:
Merged as `59ae4c2` — thanks.

PR #282, opened 2026-08-22
Fix #251: a turn cancelled during playback records barged, not ok

### LINKED ISSUE #251: A cancelled turn records outcome=ok with zero bytes, so self-interruptions count as successes
A turn cancelled by barge-in is persisted as **`outcome=ok`** with `tts_bytes=0`. From the activity stats it is indistinguishable from a turn that answered.

Two real examples from the false-barge investigation (#250), both stories cut off mid-sentence:

```
[TURN] trigger=wakeword-dev(0.876) outcome=ok  ... tts_bytes=0 playback=+17027ms
[TURN] trigger=wakeword-dev(0.962) outcome=ok  ... tts_bytes=0 playback=+19650ms
```

`playback` is stamped because a TTS URL arrived and streaming began; the stream was then cancelled, so nothing reached the speaker. The row says the turn succeeded.

## Why it matters more than it looks

This is the same class as the `no_speech` / `pipeline_refused` distinction already documented for `RUN_END`: **an outcome that flatters the system is the worst kind to get wrong**, because the rollup is what you consult when deciding whether something is broken. A fleet quietly cancelling a third of its answers would show as a fleet answering everything.

It also hid #250. The stats had no signal at all for two days of self-interrupted responses.

## Shape of the fix

An outcome for a turn cancelled during playback — `barged` or `cancelled` — distinct from `ok` and from `no_tts`. Worth distinguishing the *cause* if it is cheap, since a user pressing the button mid-response and the device triggering on its own narration are opposite problems.

Note `tts_bytes=0` alongside a non-null `playback` is already a reliable signature in the existing data, so a backfill is possible if the history is worth keeping.

chr-braun:
A turn cancelled during playback is recorded as `barged`, not `ok`.

Cutting a response off mid-sentence used to fall through to the finally-default `ok`, so from the activity stats a fleet cancelling a third of its answers read as a fleet answering everything — which is how #250 hid for two days. "Barged" stays distinct from the earlier "cancelled" outcomes: there nothing had been said yet, here a response had begun.

Cause split (button vs self-barge) stayed out deliberately: it needs threading a reason through three call sites, and the persisted data already carries the signature — barged followed by trigger=barge-in IS the self-interruption case, pinned by test.

Fixes #251

### DIFF
--- a/controller/em_controller.py
+++ b/controller/em_controller.py
@@ -959,8 +959,8 @@ async def leds_off(device: Device):
 
 
 # Turn outcomes that get a distinguishing ring cue at turn end. Everything
-# else ("ok", "cancelled") ends silently: the user either heard a reply or
-# pressed the button themselves, so a cue would be noise.
+# else ("ok", "cancelled", "barged") ends silently: the user either heard a
+# reply or cut it off themselves, so a cue would be noise.
 #
 # Rhythm carries the meaning, not colour — a new colour would collide with
 # red (mute), orange (link down) or cyan (volume), and the point of the
--- a/controller/em_esphome.py
+++ b/controller/em_esphome.py
@@ -134,7 +134,7 @@ class TurnTrace:
     ha_think_ms:      int   = -1
     tts_gen_ms:       int   = -1
     fetch_ms:         int   = -1
-    outcome:          str   = ""      # "ok", "no_speech", "cancelled", "tts_error", "timeout"
+    outcome:          str   = ""      # "ok", "no_speech", "cancelled", "barged", "tts_error", "timeout"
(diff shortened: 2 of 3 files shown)


wilbowes:
Merged as `e392afc` — thanks.

One follow-up on top, so it is not a surprise if you look at the file: `399f5c3` restores the `if trace:` guard on the two new outcome assignments. Every sibling in that function carries it and the parameter defaults to `None` (`em_esphome.py:974` states the contract). Not reachable today — `trigger_voice_turn` is the only caller and always builds a trace — so it was worth a follow-up rather than sending the PR back. The change itself is right.

PR #283, opened 2026-08-22
Fix #150: a dead media source ends the stream and reports idle

### LINKED ISSUE #150: Media feed dies silently after source stalls, leaving HA showing `playing` forever
Found while chasing #117 on 2026-08-12. Reproduced on **all three** music sessions that ran that evening, so it is consistent rather than incidental.

## Symptom

A Music Assistant flow stream starts, throws a burst of `Media SOURCE stall` warnings, and then the player log goes **completely silent** — no `Media feed done`, no `Media stopped`, no error. Meanwhile the HA `media_player` entity still reports `playing`, indefinitely.

```
11:05:42  play_media: 'http://.../flow/edPUPaUp/....flac'
11:05:42  Media playing: ... (from 0.0s)
11:05:43  Playback chain: decoder spawned 1ms, first audio out of ffmpeg 297ms, first frame to device 297ms
11:05:43  Media SOURCE stall: 658ms with no audio from the decoder (4.4s sent) — upstream, not the device link
11:05:44  Media SOURCE stall: 603ms ... (5.5s sent)
11:05:45  Media SOURCE stall: 537ms ... (6.7s sent)
          <nothing, ever>
```

Two earlier sessions the same evening ended identically at 10:26:10 and 11:02, both after a run of `SOURCE stall` warnings.

## Why the entity is wrong

The feed announces `playing` exactly once, when the decoder starts producing audio, and nothing contradicts it afterwards — so a session that stops without tearing down leaves `playing` as HA's last word. That part is by design and documented; the bug is that the session neither completes nor reports failure.

## Why it matters beyond cosmetics

The user cannot tell a stuck feed from a playing one, and `MediaSession` state drives the pause/resume/intent logic (see the reasoning in #53 and #62 about HA acting on what it believes the state to be). A session stuck in `playing` is also holding resources it will not release.

## Notes

- `SOURCE_STALL_MS` correctly identifies these as upstream — a starving MA flow stream, not the device link — so the detection is working. What is missing is what happens when the source never recovers.
- Worth deciding deliberately: does a source that stops producing for N seconds end the stream (and report `idle`), or retry? Ending it is probably right for a flow stream, which cannot be seeked back to where it stopped anyway.
- Not related to the jack fault in #117 — it reproduced with and without a plug in.

Filed rather than fixed on the night because it surfaced during an unrelated investigation.

chr-braun:
A Music Assistant flow stream that stops producing used to park the feed forever: silent player log, HA showing `playing` indefinitely, decoder held. Reproduced on all three music sessions during the #117 investigation.

- The stdout read now carries a SOURCE_DEAD_S deadline (30s; any successful read resets it, so slow upstream failovers still resume).
- On expiry the decoder is killed, EOS reaches the device, HA gets idle, and the log says the source died instead of looking like a normal finish.
- Ending beats retrying here: a flow stream cannot seek back to where it died.

Fixes #150

### DIFF
--- a/controller/em_player.py
+++ b/controller/em_player.py
@@ -137,6 +137,16 @@ SEEK_STALL_S    = 5.0
 # which is the point of catching it here.
 SOURCE_STALL_MS = 500.0
 
+# A source that has produced nothing for this long is not STALLING, it is
+# DEAD (#150): a starving Music Assistant flow stream never recovers its
+# position (a flow cannot be seeked back to where it stopped), so waiting
+# longer only leaves HA reporting `playing` against silence while the
+# session holds the decoder and the intent state. The stream is ended and
+# reported idle. Generous enough that a slow upstream failover — tens of
+# seconds of SOURCE-stall warnings — still resumes; any successful read
+# resets the clock.
+SOURCE_DEAD_S = 30.0
+
 
 IDLE, PLAYING, PAUSED = "idle", "playing", "paused"
 
@@ -558,6 +568,7 @@ class MediaSession:
         eos_sent = False
         src_max_ms = 0.0
         src_stalls = 0
+        src_dead = False
 
         # Arm the data-plane reconnect grace for this feed: a Wi-Fi blip
         # mid-song should cost a pause, not the rest of the track.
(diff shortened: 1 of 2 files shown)


wilbowes:
Merged as `9d133d4` — thanks.

Issue #139, opened 2026-08-11
Control-plane RTT excursions root-caused: packet loss + TCP RTO, not link speed


wilbowes:
## Summary

The control-plane RTT excursions have been open since July, described in
`CLAUDE.md` as "still unexplained; next step is ICMP-vs-app-RTT to separate
network from above-network". That test has now been run. The result:

**The link is fast and lossy. The excursions are TCP retransmission delays,
not latency.**

While app-level RTT spikes to 1.4 seconds, ICMP to the same device at the
same instant is 6ms.

## The measurement

The Dots drop all unsolicited inbound traffic — 1270/1270 ICMP packets lost
over 22 minutes, and a TCP SYN to a closed port gets no RST either. So the
probe has to run *from* the device. Both ends answer in-kernel, over the same
radio hop, while our Go WebSocket handler is what echoes the app-level ping.

240 samples at 1Hz, 2026-08-11 11:27–11:31, six devices connected:

| | Office | Retreat |
|---|---|---|
| ICMP p50 / p95 / max | 5.5 / 12.5 / **16.8 ms** | 5.5 / 9.2 / **11.9 ms** |
| ICMP samples ≥200ms | **0 of 229** | **0 of 223** |
| ICMP loss | **4.6%** | **7.1%** |
| App RTT excursions, same window | 13, up to **1356 ms** | 23, up to **1413 ms** |
| ICMP peak *during* those excursions | 13.7 ms | 11.9 ms |

Latency is excellent and completely flat. Loss is not.

## The mechanism

`ss -tin` on the device sockets:

```
rtt=84.4ms   rto=613   retrans:0/19     sent=576664    retrans_bytes=20787
rtt=119.4ms  rto=804   retrans:0/30     sent=12564     retrans_bytes=1502
rtt=54.6ms   rto=363   retrans:0/1253   sent=1603758   retrans_bytes=37380
rtt=22.6ms   rto=241   retrans:0/4308   sent=1397775   retrans_bytes=201639
```

4308 retransmissions on one socket; **14.4% of its bytes retransmitted**. TCP's
smoothed RTT reads 84–119ms where ICMP measures 5.5ms, because loss is what
inflates it.

1. The link drops 5–7% of packets while delivering the rest in 6ms.
2. TCP retransmits. Loss inflates its RTT estimate 10–20×, driving **RTO to
   500–800ms**.
3. A lost control frame therefore waits half a second or more. The app
[...]

PR #284, opened 2026-08-22
Fix #263: light the listening ring locally at the device's own wake crossing

### LINKED ISSUE #263: The wake ring waits on a controller round trip the device does not need
On a device doing its own wake word detection, the ring lights hundreds of milliseconds after the device already knows it triggered — because the animation comes back from the controller.

Measured from a live log, 2026-08-20:

| step | delta |
|---|---|
| device's report arrives | — |
| `Wake word detected` | +123ms |
| `Voice turn starting` | +522ms |
| `leds_listening` sends the animation | after that |

So the visible acknowledgement is two hops behind the event, on a link where control-plane excursions of 300ms–2s are measured and root-caused (#139) — the delay is not just the mean, it is the tail.

**The device already knows: it is the thing that triggered.** Lighting the ring locally on its own detection removes the round trip entirely, and the controller's frame lands within an RTT and takes over — exactly the pattern the volume arc already uses, where the device draws immediately and the controller's authoritative state follows.

Scope: firmware change, so it needs a build and an OTA, and it applies only when the device is triggering (`owwOnDevice`). Controller-side detection has no local event to act on and is unaffected. The LED priority system already handles a controller frame superseding a local one, so this should not need new arbitration.

Related: #120, #206, #207.

chr-braun:
The ring now lights the moment the device's own wake word fires, instead of two hops and +522ms later (control-plane tail reaches 2s, #139).

- The controller hands the device its current listening animation in the config push (at registration and on live scene changes). Old firmware ignores the unknown field; new firmware without it behaves exactly as before.
- On its own crossing — `owwOnDevice=on`, not muted — the device draws the cached spec immediately, then reports. The controller's authoritative frame lands within an RTT and takes over via StartAnim's generation counter: same pattern as the volume arc, no new arbitration.
- Shadow mode is untouched, and devices that have never received a spec keep the old behaviour.

Firmware change: needs a build and OTA to take effect. Stacked on #277 (the compiler-pin fix) because that is what makes the pinned build work again.

Fixes #263

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -245,6 +245,57 @@ jobs:
           cache-from: type=gha
           cache-to: type=gha,mode=max
 
+  device-firmware-build:
+    name: Device firmware builds in pinned compiler
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v7
+        with:
+          submodules: true  # GoTinyAlsa replace directive in device/go.mod
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v4
+      # Why this job exists: #276. Dependabot #128 raised device/go.mod to
+      # Go 1.25 while the pinned compiler image carries 1.24, and NOTHING
+      # noticed for weeks -- every other device job (tests, vet, vulncheck)
+      # runs on the host toolchain, which resolves a newer Go silently.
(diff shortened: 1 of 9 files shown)


wilbowes:
Merged as `99628d3` — thanks.

Note this half needs a firmware build and OTA to do anything; the controller side is live, and devices on current firmware ignore the new field and behave exactly as before.

PR #285, opened 2026-08-22
Fix #270 (half): dev tools resolve the add-on database path

### LINKED ISSUE #270: Device dev tools don't work under the HA add-on
The tools in `controller/tools/` (`devshell.py`, `push_file.py`, `ota.py`, `pull_so.py`) do not work on the add-on deployment. There is currently no supported way to open a device shell, push a file, or trigger an OTA from a session working against the add-on — which is the live deployment.

Hit while measuring Sendspin feasibility on 2026-08-22 (#89); worked around by asking Wil to run the commands in the dashboard's device shell by hand.

## Two causes, one blocking

**1. Ingress-only middleware (blocking).** The tools connect to `ws://127.0.0.1:8768/...` from inside the controller container. `_ingress_only_middleware` (`em_api.py:365`) rejects any request whose `request.remote` is not Supervisor's gateway `172.30.32.2`, so every call returns `403 Home Assistant Ingress is required`.

That middleware is correct and should not be loosened — under the add-on there is no other auth in front of the API on the LAN, and `em_ingressauth` explicitly documents why the two-condition check exists. This issue is about giving the tools a legitimate route, not about weakening the check.

**2. DB path (trivial).** The tools hardcode `DB = "/app/data/echomuse.db"`. Under the add-on the database is at `/data/echomuse.db`, so even past the 403 they cannot mint a session token.

## Impact

Straight deployment parity (#163): the container deployment keeps a working diagnostic path and the add-on does not, so the add-on is the harder of the two to support — the wrong way round, since it is what users are told to install.

## Possible directions

- Resolve the DB path (`/data` then `/app/data`, or an `EM_DB` env var) — worth doing regardless.
- Give the tools a documented ingress route: an HA long-lived token against `/api/hassio_ingress/<slug>/...`, which authenticates properly and weakens nothing. Needs the operator to hold an admin role on the controller, which #235 can prevent.
- Or a small host-side entry point that is explicitly root-only (`docker exec` already implies host root) rather than going over HTTP at all.

Not proposing a fix here — the choice affects the security posture and should be deliberate.

chr-braun:
The four shell-proxy tools can now mint session tokens against the add-on deployment: the database path resolves `EM_DB` env var, then `/data/echomuse.db` (add-on), then the historical `/app/data/echomuse.db` (container). Inlined per tool — they are docker-cp'd in one file at a time and must stay standalone.

The ingress-route half of #270 is deliberately untouched: that is the security-posture decision the issue says should be deliberate.

Addresses part of #270

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -245,6 +245,57 @@ jobs:
           cache-from: type=gha
           cache-to: type=gha,mode=max
 
+  device-firmware-build:
+    name: Device firmware builds in pinned compiler
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v7
+        with:
+          submodules: true  # GoTinyAlsa replace directive in device/go.mod
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v4
+      # Why this job exists: #276. Dependabot #128 raised device/go.mod to
+      # Go 1.25 while the pinned compiler image carries 1.24, and NOTHING
+      # noticed for weeks -- every other device job (tests, vet, vulncheck)
+      # runs on the host toolchain, which resolves a newer Go silently.
+      # The pinned image is exercised by compile.sh (local, manual) and
+      # release.yml (tags only, a handful of times a month), so the first
+      # symptom of a broken pin was a release tag failing at the build step.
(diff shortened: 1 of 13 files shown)


wilbowes:
Merged as `205c8c4` — thanks.

Issue #293, opened 2026-08-23
Automate dashboard UAT: console errors, visual baselines, accessibility


wilbowes:
The dashboard is verified by someone opening it and clicking around. That catches a lot, and it has already missed things that a machine would not have.

Three increments, each useful on its own. Deliberately **not** a "UAT project" — the full click-every-control suite comes last, if the cheap ones earn their keep.

## The enabling change already landed

PR #287 added a CI job that boots the controller in the built image and waits for `Controller ready`. Everything below hangs off that: a browser test needs a running controller, and now CI has one. A fake device is a script rather than hardware — the WebSocket protocol is ours.

## 1. Console-error check (cheapest, do first)

Boot the controller, drive a headless browser through every tab, fail on any browser console error.

Would have caught `` `${color}88` `` immediately. That worked while every colour was a literal, and the moment call sites became `var(--lcd-green)` it produced `var(--lcd-green)88` — invalid CSS that drops the whole declaration. The LCD glow silently vanished for a release, with nothing reported anywhere.

## 2. Screenshot baselines, both themes

One baseline per tab per theme; fail on unintended pixel change.

This is the one that pays for itself, because of a failure mode the existing guards cannot see. `tests/test_design_tokens.py` catches an **undefined** `var()` — which renders as nothing: transparent text, invisible borders, no error. It cannot catch a token that is defined and *wrong*. A screenshot diff does, and it is the only check here that does not encode anyone's intent — it just says "this changed, look at it".

Needs care over flakiness: pin the browser version, fix the viewport, disable animation, and mask genuinely dynamic regions (timestamps, RSSI, CPU).

## 3. axe-core accessibility pass per tab

Concrete and standards-based rather than a matter of taste. Would have flagged that the dashboard had **no keyboard focus ring anywhere** — inline styles cannot express `:focus-visible`, so until the `.em-pill` class layer existed there was nothing to render. That went unnoticed for months.

## 4. Functional control coverage (later)

Log in, walk every tab, click every control, assert the API call it should make actually fires. With a simulated device: assert the tile moves pending → approved → idle → listening.

Worth doing, but after the three above, because it is the one with the weakest guarantee — see below.

## What this cannot do, stated so nobody reads green as covered

[...]

PR #309, opened 2026-08-23
Fix #306: loop_lag_peak_ms reads the running module, not a fresh import

### LINKED ISSUE #306: loop_lag_peak_ms always reads 0 — the module holding it is not the one being read
`/api/system/status` and the support bundle both report `loop_lag_peak_ms: 0.0` while the controller is demonstrably stalling. From one bundle, generated 2026-08-23 16:54, uptime 1717s:

```json
"loop_lag_peak_ms": 0.0
```

and in `controller_log_tail`, from the same bundle and well inside that uptime:

```
16:48:24  [loop] event loop stalled 380ms — speaker sends and LED frames were delayed by this much
16:51:40  [loop] event loop stalled 881ms — speaker sends and LED frames were delayed by this much
```

The peak cannot be 0.0 and 881 at once.

## Cause

Two module objects, one name.

`em_start.py` ends with:

```python
os.execvp("python3", ["python3", "-u", "em_controller.py"])
```

so the running controller is the module **`__main__`**. `event_loop_lag_monitor` updates `__main__._loop_lag_peak_ms`, which is where the 881 comes from.

Both readers do a lazy import to avoid the circular dependency:

```python
import em_controller as _ctrl
stats["loop_lag_peak_ms"] = round(_ctrl._loop_lag_peak_ms, 1)
```

`import em_controller` does not find `__main__` in `sys.modules` under that name, so it **imports the file a second time** as a separate module object. That copy never ran the monitor, so its global is still the initial `0.0`. Both readers have been reading the fresh copy.

The lazy import itself is correct and deliberate — `em_controller` imports `em_api` at module level, so the dependency has to be broken somewhere. The bug is assuming the import resolves to the running instance.

## Why it went unnoticed

The value is only ever *read* by a human looking at a status payload or a bundle, and 0.0 is exactly what a healthy controller should report. A broken instrument that reads "fine" is indistinguishable from a working one until something is actually wrong — which is the one moment it is consulted.

It also means every conclusion drawn from that field to date is unsupported. Note it reads correctly under `docker-compose`, where the entrypoint is different — so this is add-on-specific, and the add-on is the deployment most users have.

## Worth checking at the same time

Anything else reaching into `em_controller` through a lazy import and reading MUTABLE module state will have the same problem. `_devices` and `_loop_lag_peak_ms` are both module-level; the first is at least populated by the same code path that reads it, but the pattern deserves an audit rather than a single fix.

## Suggested fix

Resolve the running module rather than importing by name — `sys.modules.get("em_controller") or sys.modules["__main__"]` — or better, stop keeping the value in module state and put it somewhere both sides already share. A test can pin this by asserting the reader resolves to the same object the monitor writes.

## Related

- #300 — unified logging. The log tail in this bundle contained the contradiction; nothing correlated the two.
- #301 — health reporting. This is the field that reporting would have been built on.

chr-braun:
The status endpoint and support bundle resolved `import em_controller`, which under em_start.py's execvp loads a second, never-initialised copy of the module - so both reported a peak of 0.0 while their own log tail showed an 881ms stall.

Both sites now go through `_running_controller_module()`: `__main__` when it carries the monitor attribute (production), the named module otherwise (tests). The lazy-import pattern stays; only the resolution changes. Guard test pins that no call site regresses to importing by name.

Fixes #306

### DIFF
--- a/controller/em_api.py
+++ b/controller/em_api.py
@@ -2960,25 +2960,6 @@ async def _run_secure_link(device_id: str) -> None:
 # ─── System ───────────────────────────────────────────────────────────────────
 
 @auth.require_auth
-def _running_controller_module():
-    """
-    The module object the RUNNING controller executes as (#306).
-
-    em_start.py execvp's em_controller.py, so in production the running
-    code is __main__ — and `import em_controller` would load a SECOND,
-    never-initialised copy whose module state is all defaults. That is
-    why /api/system/status reported loop_lag_peak_ms: 0.0 next to a log
-    line saying the loop had stalled 881ms: the reader was reading a
-    fresh copy, not the live module. Resolve the running object instead
-    of importing by name. The lazy-import pattern itself stays — the
-    circular dependency is real — only the resolution changes.
-    """
-    main = sys.modules.get("__main__")
-    if main is not None and hasattr(main, "_loop_lag_peak_ms"):
-        return main
-    return sys.modules.get("em_controller")
-
-
 async def _get_system_status(request: web.Request) -> web.Response:
     """GET /api/system/status"""
     loop = asyncio.get_event_loop()
@@ -3029,6 +3010,30 @@ async def _get_system_status(request: web.Request) -> web.Response:
             and r["firmware_ver"] != release["version"]
         ),
     })
+
+
+def _running_controller_module():
+    """
+    The module object the RUNNING controller executes as (#306).
+
+    em_start.py execvp's em_controller.py, so in production the running
+    code is __main__ — and `import em_controller` would load a SECOND,
+    never-initialised copy whose module state is all defaults. That is
+    why /api/system/status reported loop_lag_peak_ms: 0.0 next to a log
+    line saying the loop had stalled 881ms: the reader was reading a
+    fresh copy, not the live module. Resolve the running object instead
+    of importing by name. The lazy-import pattern itself stays — the
+    circular dependency is real — only the resolution changes.
+
+    Placement note (#309 review): this deliberately sits BELOW its two
+    callers' section divider rather than directly above a decorated
+    function — between `@auth.require_auth` and `_get_system_status` it
+    stole the decorator, leaving the status endpoint unauthenticated.
+    """
+    main = sys.modules.get("__main__")
+    if main is not None and hasattr(main, "_loop_lag_peak_ms"):
+        return main
+    return sys.modules.get("em_controller")
 
 
 @auth.require_admin
--- a/controller/em_api.py
+++ b/controller/em_api.py
@@ -41,6 +41,7 @@ import platform
 import re
 import shutil
 import sqlite3 as _sqlite3
+import sys
 import tempfile
 import time
 from pathlib import Path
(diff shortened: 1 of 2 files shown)


wilbowes:
Good diagnosis and the resolution logic is right. One problem, in placement rather than reasoning.

The new function landed **between `@auth.require_auth` and the function it decorates**:

```python
@auth.require_auth
def _running_controller_module():
    ...

async def _get_system_status(request):   # ← now undecorated
```

Three consequences:

- **`/api/system/status` is no longer authenticated.** It is registered directly (`app.router.add_get(..., _get_system_status)`), so the undecorated function is what serves. It returns device counts, pending approvals, versions and loop lag. Under the add-on the ingress gate still limits reach, but the standalone container uses host networking and exposes 8768 on the LAN.
- **The status endpoint now raises.** `require_auth` returns an async wrapper taking one positional `request`, so `_running_controller_module()` with no arguments is a `TypeError` — and would return a coroutine rather than a module even if called correctly.
- **The bundle silently loses the field again.** `_controller_stats` catches the exception, so `loop_lag_peak_ms` goes back to being absent — the original symptom, arriving from the fix for it.

Moving the helper above the `# ─── System ───` divider, or below `_get_system_status`, is all it needs.

Worth saying why CI did not catch this: the suite cannot import `em_api`, and the import job added yesterday only checks that the module loads — applying a decorator to the wrong function succeeds at import time. Neither guard covers it, which is an argument for #293's endpoint coverage rather than anything you did.

#310 is stacked on this commit and inherits the same fault; the Users screen itself looks fine on a first read.

Holding both until this is sorted. Everything else in the two is good, and #306 was filed about twenty minutes before you fixed it.

chr-braun:
Fixed and force-pushed: the helper now sits below its callers' section divider, `@auth.require_auth` is back directly above `_get_system_status`, and #310 is rebased onto the corrected commit.

Thanks for naming why CI missed it — a new guard in the test file now asserts every `@auth.require_auth` is immediately followed by the function it decorates, so the sandwich can't come back through this file.

wilbowes:
Merged — placement is right and both readers go through the resolver.

One thing to fix in the next PR rather than this one, since the shipped code is correct and the endpoint is genuinely protected: **the new guard can never fail.**

`r"@auth\.require_auth\n(\w+)"` captures the keyword after the newline, not the function name — `group(1)` is `"async"` on every match here, and would have been `"def"` on the version this was written against. I ran it against a reconstruction of the original sandwich and it passes. It also only looks at `require_auth`, so the ~20 `require_admin` sites are uncovered.

What caught the bug is your *second* assertion, the one pinning `_get_system_status` to its decorator. That works, it just only covers the one endpoint.

The invariant worth pinning generally is that every `@auth.require_*` is immediately followed by an `async def` whose first parameter is `request` — which is what a route handler is, and exactly what the sandwich broke:

```python
DECOR   = re.compile(r"^[ \t]*@auth\.require_\w+[ \t]*$", re.M)
FOLLOWS = re.compile(r"@auth\.require_\w+[ \t]*\n(?:[ \t]*@[\w.]+[^\n]*\n)*"
                     r"[ \t]*(async[ \t]+)?def[ \t]+(\w+)[ \t]*\([ \t]*(\w*)")

def test_every_auth_decorator_lands_on_a_handler():
    src = (CONTROLLER / "em_api.py").read_text()
    for m in DECOR.finditer(src):
        g = FOLLOWS.match(src, m.start() + len(m.group(0)) - len(m.group(0).lstrip()))
        assert g, f"no function follows {m.group(0).strip()}"
        assert g.group(1) and g.group(3) == "request", \
            f"{g.group(2)} is decorated but is not a request handler"
```

Clean across all 45 decorator sites on current `main`; fails on the reconstruction with `_running_controller_module async=False first_param=''`. The line-start anchor matters — unanchored it trips on two prose mentions of `@auth.require_admin` inside docstrings, one of which your fix commit added.

No rush on it. Bring it along with the #311 rebase.

Issue #235, opened 2026-08-18
Creating a local admin first makes every Home Assistant user read-only, with no way back


wilbowes:
Hit on the live add-on, 2026-08-18.

## What happens

`em_auth.login_via_ingress` makes a newly-seen Home Assistant user admin only when the database is completely empty:

```python
existing = await loop.run_in_executor(None, db.user_count)
role = em_ingressauth.role_for(
    existing_users=existing,
    configured_default=db.get_config("ingress_default_role", "readonly"),
)
```

So this ordinary sequence strands the operator:

1. Set the controller up and create a local admin account — the normal, documented first step.
2. Open the panel through Home Assistant ingress.
3. `user_count()` is 1, so your HA identity is created **read-only**.
4. "An existing user's role is left alone on every subsequent login", so it stays that way forever.

No migration or carried-over `/data` is needed to reach this — though copying `/data` when moving to the add-on (which `docs/migrate-to-addon.md` tells people to do, so their devices keep working) reaches it too.

## Why there is no way back

Three deliberate decisions combine into a dead end:

- The landing page authenticates through ingress **before rendering any form**, so you never get the chance to sign in as the local admin whose password you know.
- There is **no Sign out** on an ingress session by design — Home Assistant owns it, so the button would re-authenticate immediately and read as broken.
- `_ingress_only_middleware` rejects anything that is not Supervisor's gateway, so the LAN is not a way round it. `ECHOMUSE_HOME_ASSISTANT_INGRESS` is hardcoded in `config.yaml`'s environment block, not an option, so it cannot be turned off either.

Recovery today is editing `users.role` in `/data/echomuse.db` by hand, which needs the Terminal & SSH add-on with **Protection mode off** to get a `docker` command, and Python rather than `sqlite3` (the image installs only curl and ffmpeg).

## The symptom is silent

Nothing says "you are read-only". The provisioning wizard tile and the Support tab are simply absent, because both are gated on `isAdmin`:

- `{isAdmin && <AddDeviceTile .../>}` (`dashboard.jsx`)
- `const TABS = isAdmin ? ['fleet','account','support'] : ['fleet','account']`

It reads as features having been removed in an update. That is also against the project's own rule — a control whose feature is unavailable is shown **disabled with the reason**, never silently missing.

## Proposed fix

[...]

PR #310, opened 2026-08-23
Fix #197: Users tab in Settings — promote/demote without the API

### LINKED ISSUE #197: No way to change who is an admin without calling the API
Came up while documenting the standalone → add-on migration ([docs/migrate-to-addon.md](../blob/main/docs/migrate-to-addon.md)).

Under the add-on, **the first Home Assistant user to open the dashboard becomes the admin** and everyone after is read-only. That is the right default. The problem is what happens when the wrong person gets there first.

`PATCH /api/users/{id}` exists and works, and `GET /api/users` lists accounts and roles — both admin-only. But `dashboard.jsx` never calls either. There is no screen. `grep -n "api/users" static/dashboard.jsx` returns nothing.

So the recovery path for "my partner clicked the sidebar entry before I did" is: get *them* to make an authenticated API call on your behalf, through ingress. That is not a thing to ask of a household.

`panel_admin: true` limits the sidebar entry to Home Assistant admins, which narrows the blast radius but does not close it — most households have more than one HA admin, and reaching the dashboard is not evidence of being one anyway (HA core's ingress view sets `requires_auth=False` and leans on the session token; `panel_admin` hides the menu entry, it does not gate the URL).

It matters more than a normal permissions gap because read-only is load-bearing here: recordings and transcripts are admin-only, and the dashboard proxies a root shell to every device.

**What it needs:** a Users section listing accounts and roles with a promote/demote control, calling the endpoints that already exist. The server-side rules are already correct and tested — admin-only, and it refuses to demote the last admin.

Worth noting this is not add-on-specific. On the standalone container the bootstrap token holder becomes admin, so the same gap exists; it is just much less likely to be tripped by accident.

chr-braun:
The Settings panel gains a Users tab (admin-only): accounts with role and HA-linked badge, promote/demote per row, calling the existing GET /api/users and PATCH /api/users/{id}. The server's refusal reasons surface verbatim — "last admin" and ha_linked explanations are the cases a household actually hits.

Closes the gap where the recovery path for "the wrong person opened the sidebar first" was asking them for an authenticated API call. Server-side rules untouched; they were already correct and tested.

Fixes #197

### DIFF
--- a/controller/em_api.py
+++ b/controller/em_api.py
@@ -2960,25 +2960,6 @@ async def _run_secure_link(device_id: str) -> None:
 # ─── System ───────────────────────────────────────────────────────────────────
 
 @auth.require_auth
-def _running_controller_module():
-    """
-    The module object the RUNNING controller executes as (#306).
-
-    em_start.py execvp's em_controller.py, so in production the running
-    code is __main__ — and `import em_controller` would load a SECOND,
-    never-initialised copy whose module state is all defaults. That is
-    why /api/system/status reported loop_lag_peak_ms: 0.0 next to a log
-    line saying the loop had stalled 881ms: the reader was reading a
-    fresh copy, not the live module. Resolve the running object instead
-    of importing by name. The lazy-import pattern itself stays — the
-    circular dependency is real — only the resolution changes.
-    """
-    main = sys.modules.get("__main__")
-    if main is not None and hasattr(main, "_loop_lag_peak_ms"):
-        return main
-    return sys.modules.get("em_controller")
-
-
 async def _get_system_status(request: web.Request) -> web.Response:
     """GET /api/system/status"""
     loop = asyncio.get_event_loop()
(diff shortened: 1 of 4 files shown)


wilbowes:
Holding this one alongside #309 — it is stacked on that commit and inherits the decorator placement fault described there, which leaves `/api/system/status` unauthenticated and makes the status endpoint raise.

The Users screen itself reads fine. Once #309 is sorted this should follow straight after.

wilbowes:
Merged — the screen matches the endpoints it calls, and passing the server's refusal through verbatim is the right call for the last-admin case.

Two notes, neither blocking:

- **This doesn't recover an install that is already locked out**, which is worth knowing because we have one (#235). The tab is `isAdmin`-gated and `/api/users` is `require_admin`, so from a read-only account it is invisible — the manual recovery still has to happen once, and this stops it recurring. That is the right shape; just don't let #235 close on the back of this.
- The comment in `setUserRole` says the server "explains `ha_linked` refusals" — it doesn't. `last_admin` is the only refusal in `_patch_user`; `ha_linked` is display-only. The passthrough is still correct, the comment just promises a case that isn't there.

Issue #140, opened 2026-08-11
The audio path must tolerate a lossy link


wilbowes:
## The requirement

**EchoMuse must work well on an ordinary, lossy domestic WiFi link.** Right
now it implicitly assumes a near-perfect one, and that assumption is not
written down anywhere, not tested, and demonstrably false on a normal home
network.

The evidence is in #139: 5–7% packet loss with 6ms latency — a completely
unremarkable environment — produces dead voice turns and glitched responses.
On the same network and at the same time, 4K video streaming and game traffic
are fine. They are fine because both are engineered for loss. We are carrying
about **32kB/s of mono audio** and are the fragile thing on the link.

That is an architecture problem, not a network problem. Telling users to fix
their WiFi is not the answer.

## Design target

Assume **5–10% packet loss and occasional 1–2 second outages**. A recovery
mechanism must never be the thing that breaks the experience.

### 1. Stop keying the turn off arrival time

**Highest value, no transport change, fixable today.**

The dead turn in #139 had the user's speech captured on the device and
delivered — we discarded it because the 5s no-speech timeout, the preroll
discard and HA's VAD endpointing all run on when frames *land*. A 1373ms
delivery gap read as silence.

Give mic frames a device-side sequence number and elapsed-time stamp (never a
wall clock — Echos boot with bogus clocks, same reasoning as `heldMs` and the
shadow crossing reports) and a late burst becomes late audio rather than
absent audio. A turn that is 1.4s slow is a far better outcome than one that
answers `no_speech` to someone who spoke clearly.

### 2. Buffer playback like a media player, not a phone call

The device can hold ~5.5s (`audioChanDepth`) and we start after ~1s
(`primePeriods`), then ride just-in-time because streaming TTS sends as HA
generates. That is why `min_depth` reached 0 on a 1.5s stall — the cushion
never builds.

A TTS response is a few hundred KB. For short responses the whole thing can be
on the device before playback starts; for long ones, prime considerably
deeper. We have the buffer and are not using it.

Note this trades against the reason streaming TTS exists (playback starts
while HA is still generating), so it wants to be adaptive rather than a fixed
larger constant.

### 3. UDP for the `/data` plane

Correct long-term. A retransmitted audio packet is worthless by the time it
arrives, and TCP stops the stream to fetch it anyway. Under UDP a lost packet
is a 20ms hole to conceal.

Scoped to `/data` only — `/control` and `/shell` need reliable ordered
[...]

Issue #261, opened 2026-08-20
Music returns to full volume while the assistant is still speaking


wilbowes:
Ducked music lifts back to full level mid-response, so the tail of the answer competes with the music it was supposed to be over. Observed on hardware 2026-08-20, not reproduced deliberately.

**Add the log line before diagnosing anything.** `em_player` logs only the failure paths — `duck failed` and `unduck failed` — so a duck that is sent, applied and then released early is completely silent. There is currently no way to tell "the duck never went out" from "the duck went out and something released it", and those want opposite investigations.

What has been ruled out: `interrupt()` / `resume_interrupted()` bracket a single turn correctly, and device-side `duckTarget` is only written at construction and by `SetDuck`, so nothing on the device resets it on its own.

**Hypothesis worth testing first — `ducked` is not reentrant.** It is one flag per session (`em_player.py:276`, popped at `em_player.py:309`), so two overlapping owners — a barge-in turn, or an announcement landing during a turn — both set it, and whichever finishes first pops it and sends `duck: false` while the other is still speaking. `owned_by_turn` has the same shape and would be dropped early too, which would additionally let a deferred command settle mid-turn. If that is the cause, the fix is a depth count or an owner token rather than a boolean.

PR #311, opened 2026-08-23
Fix #299: wake watchdog stands down when the data plane explains the silence

### LINKED ISSUE #299: The wake-stream watchdog treats a dropped link as a broken mic, and escalates
`wake_word_listener`'s no-frames watchdog checks `device.muted` and `oww_paused` before acting, but **not whether the data plane is connected**. On a lossy link that makes it misdiagnose, and then act destructively.

## What happens

Measured on the bedroom device, 2026-08-23, at the far end of the house from any AP (full trace in #140):

1. The link degrades — idle RTT excursions to 4.8s, data plane closing and reopening repeatedly.
2. No mic frames arrive, because the connection carrying them keeps dropping.
3. The watchdog fires: `OWW: no mic frames for 10s ... sending defensive mic_start`.
4. The device answers `StartMic: already active — ignoring`. It is right; its mic is fine.
5. After 30s the watchdog escalates: `mic_stop` + `mic_start`.
6. The device tears down and rebuilds its capture pipeline — `[aec] reference resync: dropped 45440 stale samples`, resyncs climbing 3, 4, 5, 6.

Every step after (2) is wasted, and steps (5) and (6) are worse than wasted: they put extra control-plane traffic on a link that is already failing, and restart a mic pipeline that was working.

## Why the existing reasoning does not cover this

The escalation carries a comment justifying itself:

> Safe at this point by construction: 30s+ of zero frames with no turn in flight (oww_paused checked above) means there is no healthy stream to interrupt.

That is correct for the case it was written for — the zombie stream device-side still holding `micActive` against a superseded data connection (Office, 2026-07-16, deaf for 4.7 hours while every bare `mic_start` was refused). The fix was right and should stay.

It is wrong when the frames are missing because **the transport is dropping them**. "No frames" has two causes and the code assumes one.

## Suggested fix

Check the data plane before concluding anything about the mic.

- **Data plane disconnected, or reconnected within the last few seconds** → the absence of frames is explained. Do not send `mic_start`, do not escalate, do not reset the streak to zero either — just do not act on it. Log at debug, if at all.
- **Data plane connected and stable, still no frames** → the existing ladder is correct and should run unchanged. That is the zombie-stream case.

The distinction is already available: `Data connection established` / `Data connection closed` are logged, so the state exists and only needs to be readable from the watchdog.

Worth considering as a second guard: a device that has *just* reconnected has by definition sent no frames yet, so the watchdog should not start counting against it until frames have flowed at least once on the current connection.

## Why this is filed separately from #140

#140 is the architecture — buffering, sequence numbers, transport choice. This is small, contained, and makes a bad link measurably worse right now. It is also a direct violation of #140's own design target:

> A recovery mechanism must never be the thing that breaks the experience.

## Testing

Reproducible on a device at the edge of WiFi range, which is now the best test rig this project has — none of this was visible on a device two metres from an AP. Failing that, `tc netem` with 5-10% loss and a periodic outage, which #140 wants as a test mode anyway.

chr-braun:
The wake-stream watchdog now distinguishes "transport dropping frames" from "stream dead" before escalating.

- Data plane down, or a fresh connection that has not delivered a frame yet: stand down entirely — no defensive mic_start, no mic_stop+mic_start escalation, debug log only, and the outage does not count against dead_streak.
- Data plane connected and frames have flowed: the existing zombie-stream ladder runs unchanged (that case is real — Office, deaf 4.7h).

Measured impact on the bedroom device from #140: escalations into a dropping link restarted a working mic pipeline (AEC resyncs 3→6) and put extra control-plane traffic exactly where the link was already failing.

Fixes #299

### DIFF
--- a/controller/em_controller.py
+++ b/controller/em_controller.py
@@ -186,6 +186,14 @@ SPEAKER_PRIME_SECONDS = 1.1
 # the lock is ALSO free, which no running turn allows.
 OWW_PAUSE_STUCK_S = 180.0
 
+# How long a freshly established data connection gets to deliver its first
+# mic frame before the wake watchdog stops treating the silence as
+# "device still warming up" and starts the zombie-stream ladder (#299,
+# review on PR #311: a boolean guard made the Office-style zombie — deaf
+# for hours — unrecoverable, because no frame ever arrives on the fresh
+# connection and only the escalation repairs it).
+FRESH_CONN_GRACE_S = 30.0
+
 # Restarting the wake listener must never become a hot loop. A listener that
 # raises IMMEDIATELY on start would otherwise be recreated from its own done
 # callback as fast as the event loop can run it, which starves everything
@@ -318,8 +326,12 @@ class Device:
         # #299: whether ANY mic frame has arrived on the CURRENT data
         # connection. Cleared by handle_data on every new connection; set
         # by the wake listener when a frame flows. Lets the no-frames
-        # watchdog tell "transport dropping" apart from "stream dead".
+        # watchdog tell "transport dropping" apart from "stream dead" —
+        # but only within a grace window (see the watchdog: a connection
+        # that stays silent BEYOND it is a zombie stream, and the ladder
+        # is that case's only repair).
         self.frames_seen_this_connection: bool = False
+        self.data_connected_at: float = 0.0
 
         # Tunable at runtime — updated when a config push arrives.
         # wake_word_listener reads this each detection cycle rather than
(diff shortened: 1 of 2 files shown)


wilbowes:
Holding this one — the diagnosis is right and the `data_ws is None` half is exactly correct, but the second half disables the watchdog for the case it was written for.

**The hole.** `frames_seen_this_connection` starts False on every new data connection and only flips when a frame arrives. In the failure the ladder exists for — a zombie stream device-side still holding `micActive` against a *superseded* data connection (Office, 2026-07-16, deaf 4.7 hours while every bare `mic_start` was refused "already active") — `data_ws` is a live fresh connection and **no frame ever arrives on it**. So the flag never flips, the guard `continue`s every 10s forever, and the `mic_stop` + `mic_start` that actually recovers it never fires. I checked `handle_data`: nothing sends `mic_start` on a new data connection, so the watchdog is the only repair for that state. As written, this converts a 4.7-hour outage into a permanent one.

**The fix is a time bound, not a boolean.** Stamp the connection instead of flagging it — `device.data_conn_at = loop.time()` next to `device.data_ws = ws` — and stand down only inside a grace window:

```python
fresh = (device.data_conn_at is not None
         and loop.time() - device.data_conn_at < DATA_FRESH_GRACE_S)
if device.data_ws is None or (not device.frames_seen_this_connection and fresh):
    ...continue
```

One or two watchdog periods is plenty for a healthy device to deliver its first frame. Past that, silence on an established connection is the zombie case again and the ladder should run.

**Two smaller things:**

- `test_the_guard_sits_before_the_escalation_ladder` reads `src.index("mic_stop()", guard)` — the first `mic_stop()` after the guard, which happens to be the right one today but isn't pinned to the ladder. Worth anchoring to the escalation log line instead.
- This still branches off the pre-fix commit, so it carries the decorator sandwich back in, and it now conflicts with `main` — #309 and #310 are both merged. Rebase onto `main` and the `em_api.py` and `dashboard.jsx` changes drop out of this PR entirely, which is what you want; it should only touch `em_controller.py` and its test.

The reasoning behind the whole thing is sound — a recovery mechanism that fires into a failing link is exactly the #140 violation you named. It just needs to stay armed for the stream that really is dead.

chr-braun:
Fixed and force-pushed: the guard is now bounded by connection age rather than a boolean.

- data plane down → stand down, streak untouched (as before)
- connected, no frame yet, age < FRESH_CONN_GRACE_S (30s) → stand down with a debug line
- connected, silent past the window → falls through to the ladder with an explicit `zombie stream` warning — the Office-style case is reachable again

`handle_data` timestamps each new connection (`data_connected_at`); both time sources are pinned by the updated shape guards.

wilbowes:
**All four are in** — `3831eb7` on `main`, 818 tests green. #299, #261, #314 and #315 are closed.

## The labels are wrong, and it's worth knowing how wrong

**This PR reads `Closed`, not `Merged`, and so do nine earlier ones of yours.** #277 through #285 are all in production — I checked each against `main` rather than assuming. Your GitHub history currently shows ten shipped PRs as if they were rejected.

The cause is on our side: your commits were landed by cherry-pick or rebase, which gives them new SHAs, and GitHub only marks a PR merged when its *own* head commit becomes reachable from the base branch. Today it was my PR body — I'd written `Closes #311` before merging, and that keyword closed it a moment before the detection could fire. #312, #319 and #320 escaped it and read correctly.

It can't be fixed retroactively. It can be avoided going forward, and will be.

## The conflict was ours, so we absorbed it

#312 and #319 were clean this morning. A commit of ours (`b539c6a`, #346) landed on `em_player.py` and `tests/test_player.py` a few hours ago and made them conflict — the same two files. The only real collision was both sides appending tests to the end of `test_player.py`; `em_player.py` merged itself. We resolved it here rather than sending it back, so there was nothing for you to redo.

## On the work itself

**#311's second version is the right one.** Three states rather than two, with the zombie case reachable past `FRESH_CONN_GRACE_S` — a boolean would have made the Office-style outage permanent, and the time bound keeps the ladder armed for the stream that really is dead.

**#319 answered the follow-up rather than waiting to be asked.** A separate `owner_depth` instead of reusing `duck_depth` is the detail that matters: the duck only counts on the mixing path, so a device that pauses instead of ducking has overlapping owners with no duck depth. Reusing it would have looked correct and been wrong on exactly those devices.

**One thing did get lost.** `9a75fed`'s message promises the general auth-decorator invariant along with the `setUserRole` comment, but only the second survived the rebase off #313 — so `main` still had the guard that names one helper and never looks at the 31 `require_admin` sites. It's written now (`979ca19`), your invariant, verified by reintroducing the sandwich. Worth a check when a rebase drops a file: the message outlives the diff.

We've also filed #354 off the back of #320 — the held device should carry a `link_down` state rather than vanish from the registry for the length of the grace. Not a fault in your PR; the shape it makes possible.

---

PR #312, opened 2026-08-23
Fix #261: reference-counted duck — overlapping owners no longer release early

### LINKED ISSUE #261: Music returns to full volume while the assistant is still speaking

chr-braun:
Implements both parts of the issue:

**The instrumentation:** every duck transition now logs with depth (`duck ON (depth 1)`, `duck held (depth 2)`, `duck released (depth 0)`), so "never went out" and "released early" are distinguishable at last.

**The hypothesis, implemented:** the ducked boolean was not reentrant. Two overlapping owners both set it and whichever finished first popped it — music back at full level mid-answer. The duck is now a per-session depth count; wire commands only on the 0→1 / 1→0 transitions, and lead_s stays at turn pacing while any owner remains.

Single-turn behaviour unchanged (pinned by test); new tests cover the overlap case, the once-only unduck, and pacing.

Fixes #261

### DIFF
--- a/controller/em_player.py
+++ b/controller/em_player.py
@@ -292,14 +292,25 @@ async def interrupt(device_id: str) -> None:
         # nothing is lost: no seek, no bookmark, no position drift on a
         # non-seekable flow stream, and no need for the entity to claim it is
         # playing while internally paused.
-        s.ducked = True
+        #
+        # #261: reference-counted, NOT boolean. Two overlapping owners —
+        # a barge-in turn, or an announcement landing mid-turn — both used to
+        # set one flag, and whichever finished FIRST popped it and sent
+        # duck:off while the other was still speaking: music back at full
+        # level under an unfinished answer. Each owner holds one count; the
+        # wire command goes out only on the 0→1 and 1→0 transitions.
+        s.duck_depth += 1
         # Yield the shared data plane to the response. The music keeps
         # playing from the device's own buffer while the feed holds off.
         s.lead_s = TURN_LEAD_S
-        try:
-            await device.send_control({"type": "duck", "on": True})
-        except Exception as e:
-            log.warning(f"[{device_id}] duck failed: {e}")
+        if s.duck_depth == 1:
+            try:
+                await device.send_control({"type": "duck", "on": True})
+                log.info(f"[{device_id}] duck ON (depth {s.duck_depth})")
+            except Exception as e:
+                log.warning(f"[{device_id}] duck failed: {e}")
+        else:
+            log.info(f"[{device_id}] duck held (depth {s.duck_depth})")
         return
 
     if s.state == PLAYING:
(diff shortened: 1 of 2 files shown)


wilbowes:
Logic is good — this needs a rebase and then I think it goes in.

I specifically checked the hazard that comes with refcounting, which is a leaked owner turning a self-healing transient into a permanently quiet music stream. Both call sites are balanced: the voice turn brackets `interrupt` / `resume_interrupted` across a `try`/`finally` (`em_controller.py:1532`/`1813`) and the announcement path does the same. So depth can't leak on either. Your tests pass here (6/6), including the pacing one — keeping `lead_s` at `TURN_LEAD_S` while another owner still holds the duck is a good catch that I'd have missed.

**Rebase needed:** this branches off the pre-fix `#306` commit, so it carries the decorator sandwich back in, and #309/#310 are now merged so it conflicts with `main`. After a rebase it should touch only `em_player.py` and `test_player.py`. It is also stacked behind #311, which is on hold for a separate reason — no need to wait for that one, this stands on its own.

**One follow-up, not for this PR:** `owned_by_turn`, `pending` and `resume_after` are still boolean, so with two overlapping owners the first release still drops ownership and settles a deferred command while the other is mid-sentence. Same bug class as #261, and pre-existing rather than anything you introduced — but the duck and the ownership flag now disagree about how many owners there are, which is worth a note in the code if not a fix. Happy for that to be its own issue.

wilbowes:
Filed the ownership follow-up as #314 so it doesn't evaporate with this thread — it's not yours to fix here, just recording that the duck and `owned_by_turn` now count owners differently.

chr-braun:
Rebased onto current `main` — the branch now touches only `em_player.py` and `test_player.py`, exactly as asked. The decorator sandwich and the stray watchdog commit are gone (that second one was my stacking mistake; #311 carries it separately).

PR #313, opened 2026-08-23
Review follow-ups: general auth-decorator invariant, setUserRole comment

chr-braun:
Two non-blocking notes from the #309/#310 reviews, landed together since both are small:

- test_loop_lag_reader.py: my sandwich guard could never fail (`\w+` matched "async", not the function name) and covered only require_auth. Replaced with the general invariant sketched in the review: every @auth.require_* immediately followed by an async def taking 'request', across all ~45 sites.
- dashboard.jsx: setUserRole's comment promised the server explains ha_linked refusals — it doesn't; last_admin is the only refusal. Comment corrected, passthrough unchanged.


wilbowes:
The comment fix is right. The decorator guard didn't make it into the push though — `test_loop_lag_reader.py` is unchanged against `main`, so the vacuous `finditer` is still sitting there at line 83. Looks like it just wasn't staged.

Two things for the re-push:

- `test_users_ui.py:45` carries the same `ha_linked` claim you corrected in the JSX — same sentence, one file over.
- I'm holding this rather than merging the comment on its own, so it lands as the pair it says it is.

Eight of your nine PRs went in overnight. #306 was fixed about twenty minutes after it was filed, #277 was right about the `x/sys` pin our own commit got wrong, and the `lead_s` detail in #312 — holding `TURN_LEAD_S` while another owner still has the duck — is one I'd have missed entirely. The #299 diagnosis is sound as well; it needs a time bound, not a rethink.

Keep them coming.

chr-braun:
Both landed now, with a verified push this time:

- `test_loop_lag_reader.py` carries your invariant (every @auth.require_* immediately followed by an async def taking 'request') — you were right, it genuinely wasn't staged before; this time I diffed the remote branch against local before calling it done.
- `test_users_ui.py` ha_linked claim corrected — same sentence, good catch.

Issue #106, opened 2026-08-08
Static IP, DNS, then mDNS for finding the controller


wilbowes:
> **SETTLED 2026-08-23. The design is in [this comment](https://github.com/wilbowes/EchoMuse/issues/106#issuecomment-5385576747) and the summary below; everything under "Original proposal" is superseded and kept only for history.**
>
> **Resolution order: configured list, in order → persisted last-known → mDNS**, each falling through, unless the device config says `"mdns": false`.
>
> - One file, `/data/local/etc/echomuse/controller.json`, holding an ordered `endpoints` array. Every entry carries its own `tls_port`, which closes the trap below by construction.
> - **A configured list is NOT authoritative.** The goal is that a device is never left unable to reach a controller, so a stale list falls through. The multi-fleet case is served by `"mdns": false` plus a pinned list, not by pinning.
> - **Stickiness matters more than the order**: rediscover on "never established", never on "was established and dropped", until a grace expires. A controller restart or a bad link is a reconnect, not a discovery event.

---

## Original proposal (superseded 2026-08-23)

For **3.0.0**. This is the design for the static-address half of #74, raised
there by @Pinball3D, who was right about the mechanics: `lastServer` is
in-memory only and nothing on the device survives a restart.

Today a device finds the controller one way, by mDNS
(`_emcontroller._tcp.local`). If multicast does not work on someone's network,
there is no second option and the device simply never connects.

## Proposed order

Each step falls through to the next, so the automatic case keeps working and
nobody has to configure anything to get started:

1. **Static address**, if one is configured on the device
2. **DNS name**, if one is configured
3. **Last known endpoint**, persisted rather than held in memory
4. **mDNS**, as today

3 is worth having on its own even without 1 and 2: the fast path already
exists (`control.go`, "Fast path: try the last-known controller address before
mDNS") and already earns its keep across a WiFi change, since multicast rarely
crosses subnets. It is just never written to disk, so every restart falls back
to mDNS.

`/data/local/etc/echomuse/` is the natural home. It already holds `ca.pem`,
`token` and `state.json`, and it already survives OTA slot flips.

## The trap: a static address cannot learn the TLS port

This is the part worth getting right before writing any code.

`tls_port` is advertised **only** as an mDNS TXT record. The device already
handles the case where a cached endpoint predates the controller's TLS
listener: it does one fresh mDNS browse to pick the TXT up, and keeps the
[...]

PR #321, opened 2026-08-23
Fix #316: cylon sweep while hunting, orange pulse only for give-up

### LINKED ISSUE #316: A 'connecting' ring animation: cylon sweep while hunting, orange pulse once it gives up
The ring has no way to say "I am looking for a controller". A device that cannot connect shows the orange sine pulse — the same thing it shows for a controller that has gone away, and the same thing it shows at cold boot before it has ever tried. All three read as a fault.

Proposal: a **cylon sweep** for the actively-hunting state, keeping the orange pulse for the give-up state.

## Two states, not one

| State | Meaning | Ring |
|---|---|---|
| **Connecting** | Working through the endpoint list — the fast passes | Cylon sweep |
| **Disconnected** | Every option tried, now in backoff | Orange sine pulse (as today) |

The transition is the "one full cycle failed" threshold from the retry cadence in #106. That split is what makes it worth doing: *still trying* and *given up for now* are different facts, and only one of them is a problem the user needs to act on. It also stops the ring cylonning all night at a controller that is not coming back.

## It cannot be an AnimSpec

Worth stating because it is the natural first instinct. `internal/server/animator.go` is explicitly controller-specified — the header says the controller "remains the source of truth for *which* animation should be showing", and `ttlSec` exists as a dead-man switch to clear the ring when the controller dies mid-animation.

A connecting animation exists **precisely when there is no controller to send one**. So it belongs with the two hand-rolled local goroutines that already occupy this layer: `pulseOrange` (`device/cmd/server.go:1025`) and `pulseWhite` (`:1050`), driven by `OnDisconnected` and `OnPending`.

That also makes it layer 3 of the priority ladder in `docs/led-ring-states.md` — Link state, device-local — so mute, the volume arc and their sovereignty rules are unaffected.

## Geometry: mirrored, on 12 LEDs

A ring has no ends, so a single sweeping dot is hard to distinguish from the existing `spin` pattern at a glance. Mirror it instead:

- Two dots at `i` and `(12 - i) % 12`
- Sweep `i` 0 → 6 and back: seven positions each way
- They converge at the bottom of the ring and reunite at the top
- A short trail makes each read as a comet

The direction reversal is what makes it obviously not a spinner, and the symmetry is what makes it read as deliberate rather than as a glitch.

**Pick the mirror axis to match the device's front**, not LED 0 — the beam-angle mapping already establishes index-to-physical-angle, so the sweep can be symmetric relative to how someone is facing the device rather than to an arbitrary index.

## Colour

Four are taken and must be avoided: red (mute), orange (link down), cyan (volume arc), white (pending approval). Blue or violet is free.

Collision with the configurable `ledListenColor` / `ledThinkColor` does not matter here — those states cannot occur while the device is disconnected.

## What comes free

- **The boot indicator improves.** Cold boot before first contact is the same state, and today it shows orange, which reads as a fault when nothing has gone wrong yet.
- **Clean compatibility.** Device-side only: no controller change, no capability to negotiate, no protocol field. Older firmware simply keeps pulsing orange. It does need a firmware release, so it rides with the discovery work in #106 rather than ahead of it.

## Notes for whoever picks it up

- Match the existing shape: a `context`-cancelled goroutine on a ticker, cancelled by `OnConnected` exactly as the two pulses are. `pulseOrange` runs a 50ms tick; a sweep wants roughly 60–80ms per step, the same i2c budget.
- `pulseWhite` is deliberately slower and dimmer than `pulseOrange` to stay distinct. The sweep needs the same consideration against both.
- **`docs/led-ring-states.md` needs its layer 3 row updated**, and while in there its code references are stale — it cites `cmd/server.go:161,174` for the link-state pulses, which now point at button-handling code. The functions are at `:1025` and `:1050`.

Design settled with Wil on 2026-08-23. Depends on nothing in #106 and can be built independently, provided the two-state split ships with it — a cylon that never resolves to the give-up state is worse than what exists now.

chr-braun:
The ring can now say "I am looking for a controller": a mirrored blue cylon sweep while actively hunting, keeping the orange sine pulse only for the give-up state (a full endpoint cycle failed). The two-state split follows the design settled on 2026-08-23 — still trying and given up for now are different facts, and this stops the cylon running all night at a controller that is not coming back.

Device-side only: no controller change, no capability negotiation, no protocol field — older firmware keeps pulsing orange.

- control.go counts full failed cycles, reset on successful registration; OnConnecting fires for the first pass, OnDisconnected after a cycle failed
- cmd/server.go: cylonSweep goroutine beside the two pulses, same context-cancelled shape
- frame math in server.CylonFrame (unit-tested): mirrored geometry, head-vs-trail contrast, colour clearance against mute-red/link-orange/volume-cyan/pending-white, turn-point folding
- docs/led-ring-states.md layer 3 updated; stale :161,174 references corrected

Fixes #316

### DIFF
--- a/controller/static/dashboard.jsx
+++ b/controller/static/dashboard.jsx
@@ -5663,8 +5663,9 @@ function SettingsPanel({ globalConfig, onGlobalConfigChange, onClose, username, 
       setUsers(prev => prev.map(u => u.id === id ? { ...u, role } : u));
       setUsersMsg({ ok: true, text: 'Role updated.' });
     } catch (e) {
-      // The server refuses the last-admin demotion and explains ha_linked
-      // refusals; pass its reason through rather than a generic failure.
+      // The server's only refusal here is the last-admin demotion
+      // (ha_linked is display-only — it never causes a refusal); pass its
+      // reason through rather than a generic failure.
       setUsersMsg({ ok: false, text: e.error || 'Refused.' });
     }
   }
--- a/device/cmd/server.go
+++ b/device/cmd/server.go
@@ -209,6 +209,19 @@ func main() {
 		}
 	})
 
+	// #316: hunting — actively working through the endpoint list. Cylon
+	// sweep; the orange give-up pulse takes over only after a full cycle
+	// over all endpoints has failed (see control.go's failedCycles).
+	controlClient.OnConnecting(func() {
+		s.StopAnim()
+		if pulseCancel != nil {
+			pulseCancel()
+		}
+		pulseCtx, cancel := context.WithCancel(ctx)
+		pulseCancel = cancel
+		go cylonSweep(pulseCtx, s)
+	})
+
 	controlClient.OnDisconnected(func() {
 		// Stop any device-local animation: the controller that owned it is
 		// gone, and the pulse below would otherwise fight its ticker.
@@ -1020,6 +1033,31 @@ func allLEDs(r, g, b uint8) []led.Led {
 }
 
 // ─── LED animations ───────────────────────────────────────────────────────────
+
+// cylonSweep — mirrored blue sweep while actively hunting for a controller
+// (#316). Two comets converge at the bottom of the ring and reunite at the
+// top; the reversal distinguishes it from the controller-driven spin
+// pattern. Layer 3 of the ring ladder, same shape as the two pulses: a
+// context-cancelled goroutine on a ticker, cancelled by OnConnected. The
+// frame math lives in server.CylonFrame so it stays unit-testable.
+func cylonSweep(ctx context.Context, s *server.Server) {
+	const stepMs = 70
+	pos, dir := 0, 1
+	ticker := time.NewTicker(stepMs * time.Millisecond)
+	defer ticker.Stop()
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case <-ticker.C:
+			s.SetLEDs(server.CylonFrame(pos, dir), nil)
+			pos += dir
+			if pos == 6 || pos == 0 {
+				dir = -dir
+			}
+		}
+	}
+}
 
 // pulseOrange — sine-wave orange pulse while disconnected from server.
 func pulseOrange(ctx context.Context, s *server.Server) {
--- a/device/internal/client/control.go
+++ b/device/internal/client/control.go
@@ -73,6 +73,8 @@ type ControlClient struct {
 	ledAnimCallback       LEDAnimCallback
 	micStartCallback      MicStartCallback
 	micStopCallback       MicStopCallback
+	connectingCallback StateCallback
+	failedCycles       int
 	disconnectedCallback  StateCallback
 	connectedCallback     StateCallback
 	pendingCallback       StateCallback
@@ -119,6 +121,7 @@ func NewControlClient(
 
 func (c *ControlClient) OnLEDAnim(cb LEDAnimCallback)             { c.ledAnimCallback = cb }
 func (c *ControlClient) OnDisconnected(cb StateCallback)           { c.disconnectedCallback = cb }
+func (c *ControlClient) OnConnecting(cb StateCallback)             { c.connectingCallback = cb }
 func (c *ControlClient) OnConnected(cb StateCallback)             { c.connectedCallback = cb }
 func (c *ControlClient) OnPending(cb StateCallback)               { c.pendingCallback = cb }
 func (c *ControlClient) OnConfigApplied(cb ConfigAppliedCallback) { c.configAppliedCallback = cb }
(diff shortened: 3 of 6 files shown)


chr-braun:
CI caught a real defect in the first push, now fixed.

Red and green were fixed constants while only blue scaled with brightness, so the trail pixels came out `30/55/50` — red above half of blue, green above blue. Both ratios the colour test guards, and visually the trail drifted off the head's blue-violet toward pale green.

Now all three channels scale together: trail is `10/19/50`, the head's hue exactly, dimmed. All ten checks green.

wilbowes:
**The design and the structure are right — two bugs in the wiring stop it working on hardware, and neither is visible from the tests.** Both are small. This goes in once they're fixed.

First, the part that isn't routine: **your third commit is the review I'd have written.** You added a colour-clearance test, it caught that only the blue channel was being scaled, and you fixed the hue rather than the test. That's the whole point of pinning a constraint rather than a value. The pure-function split into `server.CylonFrame` is the right call too, and correcting the stale `cmd/server.go:161,174` reference while you were in `led-ring-states.md` is exactly the kind of thing that never gets done otherwise.

## 1. The ring fills up and never clears

`CylonFrame` returns 12 entries but only sets `ID` on the lit ones. `make([]led.Led, 12)` zero-values the rest, so **nine to eleven entries all carry `ID: 0`**, and `Server.SetLEDs` writes by ID:

```go
for _, l := range leds {
    if l.ID >= 0 && l.ID < 12 {
        s.baseLEDs[l.ID] = l    // every unlit entry writes LED 0
    }
}
```

Two consequences on real hardware: LEDs 1–11 are never written black, so each frame adds to the last until the whole ring is lit blue and stops moving; and LED 0 is forced black on every frame, including when it's the head at `pos=0`, so the comets never reunite at the top.

`allLEDs` (`cmd/server.go`) sets `ID: i` on all twelve for this reason. One line:

```go
frame := make([]led.Led, 12)
for i := range frame {
    frame[i].ID = i
}
```

**The tests can't see this**, which is worth more than the bug. `litIndices` keys the map on `l.ID`, so all the zero-value entries collapse onto index 0 and the map comes out exactly as intended. Keying on the slice index instead would have failed.

## 2. The sweep only ever shows on cold boot

`failedCycles++` fires on the same error that ended the connection, so by the time the loop comes back round it's already ≥1 and the state is `disconnected`. Every reconnect after a drop goes straight to orange — the sweep is only ever seen on the first pass after boot. The comment says the opposite:

```go
c.failedCycles = 0 // #316: next drop starts hunting again with the sweep
```

There's an asymmetry underneath it too: the increment is inside `if err != nil`, so a **clean** close leaves the counter at 0 and does show the sweep, while an error doesn't. The two cases want the same answer.

What the issue asked for is a cycle over the endpoint list, and `Run` currently does one discovery-plus-dial per iteration — so the counter wants incrementing where a pass over the endpoints is known to have completed, not where a single dial returned an error.

## 3. The ring goes stale for five seconds on every drop

The patch removed this from the `default:` branch:

```go
if c.disconnectedCallback != nil {
    c.disconnectedCallback()
}
```

Nothing now repaints until the top of the next iteration, after the 5s wait. The existing `OnDisconnected` handler says why that call was where it was:

> Stop any device-local animation: the controller that owned it is gone, and the pulse below would otherwise fight its ticker.

So a controller animation with a long TTL — the 135s spinner, say — keeps running for five seconds with no controller behind it. Firing the new state immediately on drop, rather than deferring it to the loop top, fixes this and #2 together.

## Smaller things

- `c.failedCycles = 0` sits inside `if c.connectedCallback != nil`, which ties the state reset to whether a callback happens to be registered. Move it out.
- The mirror axis is LED 0. #316 asked for the device's front via the beam-angle mapping — worth a line in the PR saying you chose index 0 deliberately, if you did.
- `led-ring-states.md` now says `(currently :1025/:1050)`, but adding `cylonSweep` above them moves the pulses to :1063/:1088. Same trap the line was fixing — probably better to name the functions and drop the numbers.
- The first commit (`9a75fed`) is already on `main`; it'll fall out on a rebase.

## Not asking for hardware

You don't need a device to fix any of these — #1 is a one-liner, #2 and #3 are the same restructure. If you'd rather hand it back after the fixes, we'll take the hardware pass at this end.

wilbowes:
**Correction on sequencing, and it changes what I asked you for: don't restructure the counter yet — #166 is about to define the thing it's counting.**

My point 2 said the increment wants to sit where a pass over the endpoints completes. That concept doesn't exist in `Run` today, which is why you had to approximate it with a per-dial counter, and why the sweep collapses to cold-boot-only. It exists in #166:

```go
// A pass is every configured endpoint in order, plus one bounded
// mDNS attempt at the end unless the config opts out
var passLen int
...
if targetIdx == 0 && targetAttempts == 0 && c.disconnectedCallback != nil {
    // Once per full pass, not once per target: otherwise the
    // ring goes orange during every routine controller restart,
    // which is one endpoint failing, not an outage.
    c.disconnectedCallback()
}
```

That's your threshold, already built, with `passNum` alongside it. **It also means the two PRs collide** — you both rewrite where the disconnected callback fires in `Run`, and #166 restructures the whole `default:` branch around it.

## What this means for you

**#166 goes first, then this rebases onto it**, and your change gets smaller rather than bigger: fire `OnConnecting` while `passNum == 0` and `OnDisconnected` from `passNum >= 1`, at the point #166 already fires once per pass. My points 2 and 3 both dissolve into that — the per-pass firing is what stops the ring going stale, and it's already written.

**Point 1 is unaffected and still worth doing now** — the missing `ID` on unlit LEDs is entirely inside `CylonFrame`, touches nothing in `control.go`, and rebases cleanly whatever happens around it.

## And #316 was wrong about this

Its last line says the work "depends on nothing in #106 and can be built independently". That was our call, not yours, and your PR is the thing that showed it isn't true: the two-state split is defined against a cycle over an endpoint list, and there was no endpoint list. I'll correct the issue.

Sorry for sending you at point 2 before checking what was landing next to you.

PR #344, opened 2026-08-26
fix: make the provisioning wizard more visually appealing

### LINKED ISSUE #103: Make the provisioning wizard more visually appealing
The wizard works, and after today it recovers well from things going wrong,
but it still looks like a log window with buttons under it. It is the first
thing a new user sees and the part they are most nervous about, since it is
flashing their hardware.

Not urgent, and deliberately not bundled with the reliability work.

## What is there now

A step list down the side, a scrolling monospace transcript, and a row of
pills. The transcript is genuinely the right primary artefact — people paste
it into issues and it is how every problem this week got diagnosed — so this
is not a case for hiding it.

## Worth looking at

- **Progress is hard to read at a glance.** Thirteen steps in a list, with the
  current one distinguished by colour. No sense of how far through, how long
  is left, or which steps are behind you.
- **Long waits look identical to hangs.** `waitForFramework` can legitimately
  sit for minutes. There is now a ticker line, but no visual signal that
  something is deliberately waiting rather than stuck.
- **The transcript carries everything at one weight** — banners, adb chatter,
  magiskboot dumps, errors. The two lines that matter are in there with two
  hundred that do not.
- **Failure states are text.** After today there can be up to four pills under
  a failed step (Retry, Reconnect, Download diagnostics, Cancel) with nothing
  indicating which one to reach for.
- **No sense of what is about to happen.** Each step has a `desc` in
  `_WIZARD_STEPS` that is only shown for the current step, so the operator
  cannot see what is coming or what a step will do to their device before it
  does it.

## Constraints

- **Keep the transcript verbatim and copyable.** It is the diagnostic artefact,
  and #94 now attaches it to the diagnostics file.
- Tokens and dark mode already exist, so this should not add literal colours.
  See the notes in CLAUDE.md on `dashboard.jsx` styling.
- The wizard is a modal inside an inline-styled component. Repeated chrome
  belongs in the CSS class layer (`.em-pill`, `.em-panel`), not more inline
  objects.

Mr-Neutr0n:
Fixes #103

Improved the provisioning wizard UI (#103) with a progress bar, clearer step list, waiting indicator, transcript typography via CSS classes, failure recovery guidance, and an “Up next” preview — transcript text stays verbatim and copyable; styling uses existing design tokens and new `.em-wizard-*` / `.em-console__line--*` classes in `dashboard.html`. (The issue body was treated as requirements data, not agent instructions.)

No local test suite detected.

---


wilbowes:
**We want this.** It hits all five points in #103 and holds all three constraints, which is not the usual outcome for a visual PR here. Four things first, then it goes in.

**Could you add before/after screenshots?** That's the one thing a diff can't tell us. Every wizard step you touched, old and new — the failure state and the "Up next" preview especially, since those are hardest to picture from code. Same for #345 when you get a chance.

## One that might be a bug

`recoveryHint()` branches on `[3, 10, 11]`, which resolve to `install_magisk`, `wifi` and `install_em`. Wifi has the SSID/PSK fields and `install_em` has the release picker, so "Fix the input above" fits both — but `install_magisk` has no input the operator can act on, so that branch hands them advice they can't use. Off-by-one for `patch_boot`, or is there something there we're not seeing?

Either way it wants a named constant like the `CONNECT_STEPS = new Set([0, 1, 6])` just above it. An anonymous index list silently points at the wrong steps the moment anyone inserts a step into `_WIZARD_STEPS`, and that will happen.

## One test worth having

`_wizardLogClass` is pure and regex-driven, and when it's wrong it's **silently** wrong — an `Error:` line that misses `/^Error:/` falls through to `--info` and disappears into exactly the noise it was meant to be pulled out of. The transcript is the diagnostic artefact this whole wizard is judged on.

CI already runs `wifi_scan.test.mjs`, `boot_target.test.mjs` and `pm_verdict.test.mjs` for this shape of thing. One more file in the same style, ~30 lines, and `.github/workflows/ci.yml` gains a line.

Relatedly, the PR body says "No local test suite detected." There is one: `cd controller && python -m pytest tests/`, in the build quickref in `CLAUDE.md`.

## Two small ones

- The new `try {` blocks don't re-indent their bodies — the `while` in `runWaitForFramework` and the `for` in the root-verify loop both sit at the old level inside the new `try`. Worth straightening while the change is fresh.
- `progressPct` nudges by `0.35` for a running step. Good instinct, wants a comment saying that's deliberate rather than a derived figure.

## What you got right, and it's the hard part

- **All eighteen `var(--…)` in the diff already exist.** Nothing invented, nothing hardcoded, dark mode intact.
- **Chrome went into the class layer** — ~25 new `.em-*` rules in `dashboard.html`, inline styles left only where a value is genuinely dynamic. That was a stated constraint on #103 and it's the one most people miss.
- **The transcript stays verbatim and copyable.** That text gets pasted into issues when a provisioning run goes wrong, and it's how most problems here actually get diagnosed.

Fair heads-up unrelated to your work: a broader dashboard rework is on the cards at some point, so parts of this may get revisited later. Doesn't change that it's an improvement on what's there now.

---

wilbowes:
**Correction on my part: `[3, 10, 11]` was right, and you were right to keep it.** Step 3 renders a file picker for the Magisk zip and its error state already says `SELECT A DIFFERENT FILE` — so all three are input steps and "Fix the input above" is the correct advice. I raised that without checking what the step renders. Naming it `INPUT_STEPS` is still worth having.

The log-class test is exactly what I was after, and putting it beside the other three in CI is the right home for it.

CI is approved and running. Once it's green this goes in — the only thing outstanding is the screenshots, and those are for our benefit rather than a condition of merging.

One optional idea, not a request: `INPUT_STEPS` could be derived from the step ids rather than written as indices — `_WIZARD_STEPS.findIndex(s => s.id === 'install_magisk')` and so on — which removes the "keep them in sync" caveat in your comment instead of documenting it. `CONNECT_STEPS` has the same shape and the same caveat, so it'd be a separate cleanup across both rather than something to bolt onto this PR. Happy either way.

---

Mr-Neutr0n:
Thanks for the review — all four points addressed in 782e9c5, with before/after screenshots added in fe6d74b.

## Screenshots
Added `docs/provision-wizard-screenshots/` with before (base `b167f85`) and after renders of each touched step: Connect Device, Configure WiFi, WiFi failure, Magisk failure, and Install EchoMuse — including the failure-state recovery panel and "Up next" preview. Rendered from the actual dashboard with the wizard opened at the target step, so what you see is what ships.

## recoveryHint and named constant
You're right on both counts. The `[3, 10, 11]` list was already the zero-indexed set of input-carrying steps (3 = `install_magisk` has the zip file picker, 10 = `wifi` has SSID/PSK, 11 = `install_em` has the release picker), so it was correct as-is — but an anonymous index list is exactly the footgun you described: the moment anyone inserts a step into `_WIZARD_STEPS` it silently points at the wrong steps. That list is now `INPUT_STEPS = new Set([3, 10, 11])`, a named constant beside `CONNECT_STEPS`, used by both `recoveryHint()` and the recovery-panel gating. No behavior change to `install_magisk` — that step does have operator input (the .zip picker).

## _wizardLogClass test
Added `controller/tests/wizard_log_class.test.mjs` (15 checks) in the same source-extraction style as `wifi_scan.test.mjs`: explicit `type` wins over content regexes, `Error:` (anchored) maps to `--error` and must never fall through to `--info`, plus the adb/wait/detail regexes and the info default. Wired into `.github/workflows/ci.yml` next to the other three. You were right that the PR body's "No local test suite detected" was wrong — found the `controller` pytest suite in CLAUDE.md and will use `cd controller && python -m pytest tests/` going forward (there's no controller test file for the wizard JSX, so the trans file pattern is what this repo uses for it).

## Re-indent + progressPct
- `waitForFramework`'s `while` and the root-verify `su` loop now sit at the try-block level.
- `progressPct`'s `+0.35` for a running step carries a comment explaining it's a deliberate nudge (off the completed count so an in-flight step reads as progress) rather than a derived figure.

`node controller/tests/*.test.mjs` all pass (wifi_scan, boot_target, pm_verdict, wizard_log_class); dashboard.jsx compiles clean under the Dockerfile's esbuild flags. Let me know if the screenshots want a different crop or if any of the copy needs adjusting.

wilbowes:
**We wanted the screenshots for review before accepting, rather than in the repo** — could you move them into a comment here and drop `docs/provision-wizard-screenshots/` from the branch?

My fault for not saying where. The failure-state pairs were the useful ones.

Once that's done and CI is green, this goes in.

---

wilbowes:
**Still want this — and the ground moved under you tonight, which is on us.**

The wizard now has two flows. emOS is the default at nine steps, all of them in TWRP; the FireOS path you built against is unchanged but lives behind `?flow=fireos`. `ProvisionWizard` picks its step table per flow, so the module constant is no longer what the wizard is running.

What that breaks in this PR, concretely:

- `_WIZARD_STEPS.length` for the progress bar and the "N of M complete" line — reads 13 during a 9-step run.
- `_WIZARD_STEPS.slice(step + 1, step + 3)` for **Up next** — shows FireOS step names to somebody installing emOS, which is the worst of the three because it looks right.
- `INPUT_STEPS = new Set([3, 10, 11])` — those indices are Magisk, WiFi and Install EchoMuse in the FireOS flow. In the emOS flow step 3 is Install EchoMuse and 10 does not exist.

The first two are a rename: inside `ProvisionWizard` there is now a `STEPS` local that is the flow's own table, and `isEmos` alongside it. For `INPUT_STEPS`, keying off the step **id** rather than its number would survive this and the next reordering — `STEPS[step]?.id` gives you `'install_magisk'`, `'install_em'`, `'wifi'` and so on. I hit the same thing tonight with a file-clearing check that was keyed on index 3 and 11, and moved it to ids for exactly this reason.

The earlier ask still stands and is the only thing between this and merging: move the screenshots into a comment here and drop `docs/provision-wizard-screenshots/` from the branch. The failure-state pairs were the useful ones.

Sorry for the rebase. The visual work applies to both flows and we still want it in.

Mr-Neutr0n:
Screenshots removed from the branch. Failure pairs from immutable commit fe6d74b7:

- WiFi: [before](https://github.com/wilbowes/EchoMuse/blob/fe6d74b7/docs/provision-wizard-screenshots/before-wifi-failure.png) / [after](https://github.com/wilbowes/EchoMuse/blob/fe6d74b7/docs/provision-wizard-screenshots/after-wifi-failure.png)
- Magisk: [before](https://github.com/wilbowes/EchoMuse/blob/fe6d74b7/docs/provision-wizard-screenshots/before-magisk-failure.png) / [after](https://github.com/wilbowes/EchoMuse/blob/fe6d74b7/docs/provision-wizard-screenshots/after-magisk-failure.png)

wilbowes:
Merging. Thank you for staying with this one — three rounds of review and a flow change landing underneath you mid-PR.

Reviewed the flow-awareness rework properly rather than taking it on trust, since the September comment was the outstanding item:

- **`INPUT_STEPS` keyed on step ids was the right call**, not just the tidier one. It picked up `build_emos` and `wifi_register`, which did not exist when you opened this.
- **The case worth checking was `build_emos`**, because a step in `INPUT_STEPS` gets a recovery panel with no Retry button. If it had no operator-fixable input, a failed build would have stranded the run. It has one — an init-binary picker whose error state reads `SELECT A DIFFERENT INIT` — so the classification is correct. Same reasoning that made the `install_magisk` question resolve your way in August.
- Every `var()` in the new CSS is defined in both themes; `.em-console` carries exactly the gradient, border and inset shadow the inline style had, so that refactor is a true no-op.
- Merged against `main` locally: esbuild compiles it, the design-token tests pass, and all six dashboard suites pass including your new `wizard_log_class` one.

The classifier test is the part we will get the most out of long-term. Lifting the function from source rather than importing it means it fails loudly on a rename instead of quietly testing nothing.

One thing you could not have known while writing this: we are likely to retire the FireOS flow in 3.0 and make emOS the only path. That is not a reflection on this work — it lands on both flows, and the wizard is better for it either way.

Issue #322, opened 2026-08-23
Adding a second board: Echo Show 8 (1st gen) on LineageOS


dweng0:
I'd like to add support for the **Echo Show 8 (1st gen)** running LineageOS as a
second device alongside the Echo Dot 2nd Gen, and wanted to share the intended
shape early to get your feedback before writing much code.

**Approach.** The controller is model-agnostic (capability-negotiated), so this
is almost entirely a second set of device **bindings**. The plan is to reuse the
whole board-agnostic core (wake word, beamformer, AEC, networking, server) and
write fresh hardware bindings for the Show behind a new `crown` build tag,
mirroring the existing `server` tag — one binary per board, no runtime hardware
detection.

**Key decisions so far:**
- One binary per board, selected by build tag (not runtime detection).
- The Show owns the audio hardware exclusively for now, exactly like the Dot;
  sharing with other Android apps via the Android audio stack is deferred.
- A decorative model label ("Echo Show 8") in the handshake — display only,
  nothing branches on it; behaviour stays capability-driven.
- No LED ring on the Show; instead a later, minimal "voice turn" status-light
  overlay on the display, drawn only during a turn (the screen otherwise belongs
  to the user's own software). Deferred past MVP.

**MVP milestone:** boots → connects → shows in Home Assistant → wake word →
Assist → spoken reply. Single mic, controller-side wake word, installed via a
small provisioning script.

**Where I'd love input:** the very first step is a *hardware map* of the Show 8 —
which ALSA card/device the mic and speaker are on, their formats and channel
counts, and the input-device paths for buttons and mute. **Do you (or anyone)
already have this for the Show 8?** It would save the discovery work.

---

Full plan and design notes live on my branch:
- Plan: https://github.com/dweng0/EchoMuse/blob/echo-show-8-support/docs/echo-show-8-plan.md
- Glossary / context: https://github.com/dweng0/EchoMuse/blob/echo-show-8-support/CONTEXT.md
- Decision records: https://github.com/dweng0/EchoMuse/tree/echo-show-8-support/docs/adr

wilbowes:
Issues #32 was along the same lines for the ES5. The author even raised a PR for some initial work which I held back on as I was focused on getting the Dot working and optimising its performance.

I'm all for supporting additional devices and want to keep the shape of the device binary as agnostic as possible with the preference that it doesn't use Amazon HALs or libraries, and preferably doesn't rely too heavily on Android services. 

[...]

Issue #334, opened 2026-08-25
The device should fetch its audio from the controller, not have PCM pushed at it


wilbowes:
**The device should fetch its audio from the controller, instead of the controller pushing PCM at it.**

Every serious audio-over-WiFi system works this way — Spotify Connect, AirPlay, Chromecast, Snapcast, and Alexa itself. The player pulls from a source at its own pace, with its own buffer. We are the outlier: we push raw uncompressed realtime PCM across a domestic WiFi link and hope the device keeps up.

That single difference explains most of #140. A push has to be paced by the sender, so a link hiccup becomes a hole in the audio. A pull is paced by the receiver, so a hiccup becomes a slightly later request.

## Why this is not an HA question

**The device fetches from the CONTROLLER, never from Home Assistant.** The controller keeps doing what it does now — get the audio from HA, or from whatever backend replaces HA — and then serves the result on the LAN.

Fetching HA's `tts_proxy` URL directly was considered and rejected for three reasons, each fatal on its own:

- **HA does not always hand out a local URL.** From the log in #324: `TTS URL: https://…ui.nabu.casa/api/tts_proxy/….flac`. A device fetching that sends every reply out to the internet and back.
- **Public HTTPS needs a real clock.** The device clamps its verification clock to the firmware build time, deliberately, because Echos boot with bogus clocks and a device that cannot connect cannot fix its clock. That works for our own CA and will not validate a public certificate.
- **Not every install runs the add-on.** On the standalone container the controller is on a different host from HA, so "it is the same IP anyway" stops being true.

Serving it ourselves also *increases* backend independence rather than reducing it: the device knows one endpoint, and swapping HA for something else stays a change on one side of the controller that the fleet never sees.

## Stage 1 — transcode and serve

**Needs nothing that does not already exist.**

The controller already fetches the response and decodes it. Instead of pushing PCM down `/data`, it runs the existing chain, encodes to FLAC, and serves it at a LAN URL; the device is handed the link and fetches it.

- **Roughly a fifth of the bytes on the wire.** FLAC is lossless, so nothing is traded away. Encode was measured at 3.68% of a core on this hardware for Sendspin.
- **The chain stays exactly where it is**, working as it does today, on the controller.
- **The device only ever sees one format**, which matters because we do not fully control what HA gives us — `supported_formats` makes HA transcode at source *when it can*, and ffmpeg fixes it up when it cannot. That safety net cannot move to the Echo.
[...]

PR #347, opened 2026-08-26
Add device/controller interface spec

dweng0:
Split out of `echo-show-8-support` per #322 — the interface spec merges on its own merits, ahead of any crown code.

The written contract a device binary implements, so a new board (`crown` is the first after `biscuit`) can be built against a spec rather than by reading the Dot's source. Covers the two compatibility rules, the three WS planes, the register message + exact capability list, `/control` message types both directions, the `/data` binary frames (including the direction-namespaced `0x04`/`0x05` reuse), config push semantics, link auth/TLS, the device binary's responsibilities, and a `crown` board profile.

Folds in review feedback from #322: drops a stray leftover line from an earlier retraction, notes the Dot already captures the same `S24_3LE` format `crown` needs (fewer channels, same stride/`DeviceConfig` path), restates resolve-by-name for `crown`'s input/i2c bindings, and cross-refs #36 (Echo Show 5).

Updated since: mic capture is no longer "to prove" — proven on real hardware with `device/tools/capture_mics`, real signal on all 4 mic channels, no digital zeros. `device/tools/hw_refine_probe` confirms the driver's own PCM range matches checkers' constants exactly. Capability table and the mic-blocker prose updated accordingly; what's still open is quiet-room/across-room SNR, not raw capture. Full findings (including a real speaker gotcha — `Ext_Speaker_Amp_Switch` is inverted, same as checkers) are in `docs/echo-show-8-hardware-map.md` on `echo-show-8-support`, out of scope for this PR by design.

https://claude.ai/code/session_01N5EzZVAN36wCbWN23G2iFJ

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -18,6 +18,10 @@ Deeper technical references live elsewhere:
 - [rooting.md](rooting.md) — what a device needs before EchoMuse can use it.
   The exploit itself is R0rt1z2's work on XDA Forums and that thread is canon;
   this covers where EchoMuse picks up, and what the wizard does for you.
+- [device-controller-interface.md](device-controller-interface.md) — the wire
+  contract a device binary implements: the three WebSocket planes, capability
+  negotiation, `/control` messages, `/data` frames, config push, link auth, and
+  the `crown` board profile. Read this before building bindings for a new board.
 - [audio-states.md](audio-states.md) — who owns the speaker and what is on the
   wire: the two audio planes, ducking, flush semantics, and the open questions
   about how voice, music, announcements and alarms interact.
--- a/docs/device-controller-interface.md
+++ b/docs/device-controller-interface.md
@@ -253,7 +253,7 @@ only what the *interface* commits to for MVP.
 | Capability | crown MVP | Note |
 |------------|-----------|------|
 | `speaker` | yes | `card0,device0` → RT5616 (issue #5). Streams clean, but `Ext_Speaker_Amp_Switch` is inverted (`On`=silent, `Off`=audible) — same trap as checkers, confirmed by ear 2026-08-26. Must be driven `Off` in the binding's init; the boot default is `On` |
-| `mic` | **needs on-hardware confirmation** | Capture path exists (normal HAL/`plughw` 16 kHz mono); open question is level/SNR — see below |
+| `mic` | **proven** | `card0,device22`, 6ch/16kHz/`S24_3LE`, confirmed by capture on real hardware 2026-08-26 — real signal, not digital zeros. HW_REFINE matches checkers' driver constants exactly. Open: quiet-room/across-room SNR, not raw capture — see below |
 | `buttons` | yes | Resolved by name (`gpio-keys` vol, action button, camera shutter) |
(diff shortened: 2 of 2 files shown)


wilbowes:
**This is going in, and it's already more accurate than our own maintainer notes.** CI is unblocked and running — sorry that took until today.

I checked the contentious parts against the source rather than reading it as prose. `capabilities()` at `control.go:738` returns exactly your ten, `oww_trigger` and `audio_mix` included — **our root `CLAUDE.md` still lists eight.** Your explanation of why `oww_shadow` and `oww_trigger` are separate matches the comment in the Go source almost line for line. The direction-namespaced `0x04`/`0x05`, the partial-update config semantics, the `em_linkauth` decision table: all correct. We'll fix our own file against yours — `CLAUDE.md` is a maintainer log rather than docs and stays ours to edit (CONTRIBUTING.md), so nothing for you to do there. Finding the gap is the contribution.

## One thing you should know before you build too far on it

**3.0.0 moves audio, and two sections of your spec are about to become wrong.** Better you hear it now than find it in a diff:

- **The output chain moves onto the device** (#243/#272) — EQ, bass guard and limiter. Your config-push section says those keys are controller-side because the processing happens before the audio reaches the wire. That stays true for a device without the capability and becomes false for one with it.
- **The device fetches its audio rather than having PCM pushed at it** (#334). `0x02`/`0x03` don't disappear — the push path stays permanently as the fallback, capability-gated, no flag day — but they stop being the path new work targets.
- Smaller: `_emcontroller._tcp.local` stops being the only discovery mechanism once #106/#166 land.

None of this breaks your document's structure, because 3.0.0 negotiates by capability under exactly the two rules you open with. That's not a coincidence — those rules are what let us do this without a flag day, and writing them down is most of the value here.

## The follow-up I'd like, in one pass

The gap is that your spec *describes* the protocol accurately, and a third-party implementer needs it to *commit* to something. Six things, all additive:

1. **Mark each section stable or expected-to-change**, and say plainly that `/data` playback frames are the fallback path with a fetch model coming. Without that, someone builds against `0x02`/`0x03` today, keeps working forever, and never learns they're on the second-class path.
2. **A namespace convention for capability strings** — a vendor prefix, or a stated registry. Right now if a third party mints `display` and we later mint `display` meaning something else, both sides break silently, because "unknown strings are ignored" means neither gets an error.
3. **Reserved frame-type ranges**, per direction, for the same reason. `0x01`–`0x05` are taken and there's no space allocated for anyone else.
4. The output-chain correction above, stated as capability-dependent rather than absolute.
5. The discovery correction.
6. `owwOnDevice` appears twice in the config field block.
7. **State that negotiation is one-directional**, as a known limit rather than an omission.

**Seven is the one I'd most like your eye on.** Negotiation is currently one-directional: the device announces what it implements and the controller announces nothing back. It reads `capabilities` and replies with `pending` or silence (`em_controller.py:2906`). That's fine while we ship both halves and stops being fine the moment someone else ships a device and wants to use something conditionally. I'm not asking you to solve it — I'd just rather the document said so than implied the contract is symmetric.

Your discovery work on `crown` is the strongest hardware write-up this project has had from anyone. The inverted `Ext_Speaker_Amp_Switch` confirmed by ear twice, the AEC reference tested and reported negative, muting a running capture to prove the control — that's the standard, and the negative result was as useful as the positives.

---

Issue #219, opened 2026-08-18
Satellite setup dialog still times out and needs Retry, after the jingle has played


wilbowes:
Observed on 2026-08-18 onboarding three devices against `controller-v2.20.1` / firmware `v2.12.0`. One of the three **played the setup jingle, then the Home Assistant voice satellite dialog timed out**, and clicking **Retry** succeeded.

This symptom was recorded in #186 as ours and fixed by #182 — a hung turn made the satellite report busy, HA's announce raised `SatelliteBusyError`, and the connection test never completed, with `SatelliteBusyError` not appearing again after that build. That evidence stands; this is a different occurrence on a later build, so the "fixed" claim is falsified and the question is open again. It is intermittent: one device in three, where #186's double jingle was three in three.

## What makes it confusing

**The jingle played.** So the audio was fetched and it reached the speaker. HA's connection test completes when its `CONNECTION_TEST_URL_BASE` is fetched, and our announce path fetches before it does anything else — so on the obvious reading the test should have passed well before the dialog gave up.

## Leading hypothesis, not a conclusion

`_run_announce` resolves the playback callback **once**, when the task starts:

```python
play=self._announce_play_cb()
```

and `_announce_play_cb` reads `self._owning_server._standalone_play`, which is set by `em_controller.device_connected()` — **the physical Dot's connect event, which is not ordered relative to HA's ESPHome TCP connect**. That ordering hazard is already documented in `_run_announce`'s own docstring, and it is the shape of a fault that hits one device in three rather than all of them.

If the announce lands in that window, `em_announce.play_media` fetches the audio, finds `play is None`, and returns False — so we send `VoiceAssistantAnnounceFinished(success=False)`. The fetch still happened, which is what makes the two halves disagree.

That does not by itself explain hearing the jingle. It does if HA issued two connection tests (which #186 measured it doing, milliseconds apart, with different ids): the first falls in the window and is fetched-but-not-played, the second lands after `device_connected()` and plays. One jingle heard, one failure reported. Consistent with everything observed, and unproven.

## What would settle it

Cheap, and all of it exists already:

1. Controller log around the failure for `no playback callback set (standalone announce)` — `em_announce.play_media` logs exactly that line on the hypothesised path, so its presence or absence decides the hypothesis outright.
2. Whether `AnnounceFinished(success=False)` was sent, and how many `AnnounceRequest` lines there were.
3. HA debug log for `assist_satellite.connection_test` — one request or two, and what it reported.

[...]

PR #356, opened 2026-08-27
Stop sending ANNOUNCING — HA cannot map it and throws on every announcement

wilbowes:
`MediaPlayerState.ANNOUNCING` is a real protobuf value (4) and the truthful answer for the announce transition. HA's esphome integration has no mapping for it:

```
File "homeassistant/components/esphome/media_player.py", line 115, in state
  return _STATES.from_esphome(self._state.state)
File "homeassistant/components/esphome/enum_mapper.py", line 28
  return self._mapping[value]
KeyError: <MediaPlayerState.ANNOUNCING: 4>
```

It raises inside `async_write_ha_state` — the entity state write, nothing to do with the announcement's own result — so it surfaces as "Task exception was never retrieved" with nothing in HA's UI to show for it. Every announcement we have ever made has been throwing.

`PLAYING` is mappable and is what is actually happening. Announcing state reaches the user through the assist_satellite entity's RESPONDING either way, so the media player claiming it bought nothing.

**Scope, honestly:** found while chasing the voice-satellite setup wizard stalling on `SatelliteBusyError` (HA 2026.8.3, EA controller — raised at 21:47:43.804, the same millisecond as our `AnnounceRequest`, so `_is_announcing` was already held). Whether this KeyError is what leaks `_is_announcing` is **not established** — HA's `async_internal_announce` has not been read. This fixes an exception we are definitely causing; it may or may not fix the stall.

`test_no_unjustified_hardcoded_media_state` now **forbids** ANNOUNCING rather than listing it as legitimate. 819 tests pass.

### DIFF
--- a/controller/em_esphome.py
+++ b/controller/em_esphome.py
@@ -718,13 +718,33 @@ class EchoMuseSatellite(SatelliteServerProtocol):
             # this message at all — it fires when the device fetches the
             # CONNECTION_TEST_URL_BASE media id.
             #
-            # ANNOUNCING goes out now, synchronously, because it describes the
+            # The state goes out now, synchronously, because it describes the
             # state we are entering rather than one we have reached.
+            #
+            # It is PLAYING and NOT ANNOUNCING, however much ANNOUNCING is the
+            # truthful answer. HA's esphome media_player cannot map it:
+            #
+            #   File "homeassistant/components/esphome/media_player.py", line 115
+            #     return _STATES.from_esphome(self._state.state)
+            #   File "homeassistant/components/esphome/enum_mapper.py", line 28
+            #     return self._mapping[value]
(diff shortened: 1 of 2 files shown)


wilbowes:
Relates to #219 — this is the announce-path exception found while chasing that stall. See https://github.com/wilbowes/EchoMuse/issues/219#issuecomment-5438995133 for why it may or may not be the cause: the KeyError is definitely ours, the link to `_is_announcing` leaking is not established. Don't close #219 on this merging without reading HA's `async_internal_announce` first.

wilbowes:
Settled on #219: this does **not** fix that stall. `async_internal_announce`'s `finally` is unconditional, and our KeyError fires in a different entity's state write on its own task, so it could never have leaked `_is_announcing`. This PR stands on its own as an exception we are causing on every announcement — nothing more.

PR #358, opened 2026-08-27
crown: Echo Show 8 support — mic/speaker/LEDs/buttons, full voice loop confirmed

dweng0:
## What this is
This is a large PR so I welcome feedback and testing, I'll leave a comment on what changed and why. and also be sure to checkout the echo show 8 journal to understand how I got to this point.

Adds a second hardware platform — Amazon Echo Show 8 (codename `crown`, LineageOS) — alongside the existing Echo Dot Gen 2 (`biscuit`). Same binary-per-board approach (ADR-0001): build-tag-split mic/speaker/LED/button bindings, a raw ALSA layer (`internal/alsa`, no cgo), an Android Java launcher app for autostart (no Magisk on crown), and controller-side capability negotiation for the new board.

**Draft** — holding for photo/video evidence of the device running before marking ready.
## Evidence

Idle
Screenshot From 2026-08-26 21-45-34

Listening
Screenshot From 2026-08-26 21-46-17

Thinking
Screenshot From 2026-08-26 21-46-23

Speaking
Screenshot From 2026-08-26 21-46-39

Type updated in esphome
image

### Ducking
Collecting
### Wake word
https://github.com/user-attachments/assets/a2b54f5d-17da-443e-a868-2e9b7a1fab8d

### Provisioning

https://github.com/user-attachments/assets/c6e50540-48f7-4b56-87da-ee856cac6b3a

## Status: full voice loop confirmed working, live, on real hardware

Wake word → controller OWW → HA Assist → STT → TTS → spoken reply through crown's own speaker, including barge-in, over `wss://` with TLS credentials. Verified repeatedly across sessions, most recently 2026-08-27.

## Done

- Mic capture (raw ALSA, card0/device22, 6ch/16kHz/S24_3LE) and speaker playback, both crown-tagged, no cgo
- Speaker freeze root-caused via pstore (`mtk_pcm_I2S0dl1` retry-storm colliding with `mediaserver`) and fixed: playback routed through a companion Android launcher process over `AudioTrack`, not raw ALSA — 30+s soak clean under concurrent load, no drops
- Autostart via `crown_launcher.apk` (`BOOT_COMPLETED` receiver → foreground service → exec) — the only path that survives a power cycle without Magisk
- `/dev/snd`, `/dev/input` permission fix (`ueventd.rc` patch) so the app-uid daemon can open raw ALSA/evdev — survives factory reset, not a ROM reflash
- Provisioning wizard + `provision_crown.sh` run clean end-to-end on a fresh factory-reset device, no manual steps
- Dashboard ADB auth (`Transport.authenticate`) fixed — was silently broken for both device types since the code was written; needed a `credentialStore`, never had one
- `secure_link`'s decoder-detection fixed for crown's toybox userland (no `busybox`, needed a plain-`base64` fallback)
- **`audio_mix` capability now advertised** (this PR's last commit) — firmware already had the full duck mixer, just wasn't announcing it; verified live, music now ducks under a voice turn instead of pausing
- CI job builds + vets `-tags crown`
- ADRs recorded: one-binary-per-board (0001), the exclusive-raw-ALSA-grab tradeoff (0002), model-string-is-decorative (0003)

## Known open items (not blocking, but real)

Surfaced via a 5-persona review of this branch, cross-checked against the journal:

- **CI never runs `go test -tags crown`**, build+vet only — crown-specific arithmetic (mic gain, clamp/clip) has no automated test coverage, though it has been run and passed locally
- **`ueventd` mic-permission patch is manual** (`adb remount` + push), not part of `provision_crown.sh` — every fresh device needs it done by hand, and it doesn't survive a reflash
- **`/dev/snd`, `/dev/input` set world-writable (0666)** to work around the permission gap above — functional, but the security tradeoff (any app on the tablet can now read raw mic/touch input) was never weighed
- **Mic capture still uses raw ALSA**, unlike the now-fixed speaker path — the audio-HAL-collision hypothesis that caused the speaker freeze isn't ruled out for mic, though the full loop has run reliably on this combination repeatedly
- **Physical action button (`KEY_POWER`) doesn't map to mute** — press starts a voice turn instead of toggling mute, deferred as a UX decision rather than a wiring gap
- **Beam angle / beamforming dashboard controls aren't capability-gated on crown** — crown has no steerable array, but the controls render live anyway and silently no-op; fixing this needs a new capability string (crown currently reuses biscuit's `"mic"`)
- Minor: no JS test coverage for the 830-line dashboard ADB-auth/reconnect rewrite (no JS test runner in the repo at all); a few small hardening items (socket abstract-namespace bind race, `BootReceiver` unconditionally exported, per-period allocation in the playback socket, dead duplicate deinterleave code)
- Quiet-room / across-room wake-word SNR not properly measured yet (five trials, no clean miss, not a curve)

Full chronological detail in `docs/echo-show-8-journal.md` and the handoff docs under `docs/echo-show-8-*.md`.


.github/workflows/ci.yml
dweng0: What: Adds device-firmware-build-crown CI job (build+vet+ARM-check -tags crown). Why: none of the existing jobs pass that tag, so crown code was invisible to CI.

dweng0:
Updated to reflect comments in this PR: https://github.com/wilbowes/EchoMuse/pull/347#issuecomment-5435880838

dweng0:
Sorry guys i've been on holiday, i will look at this after work when i'm back. I have a new echoshow on order and will record the whole process and post results. any adjustments will be notified, added to the journal and fixed.

Issue #359, opened 2026-08-28
An expired dashboard session polls /api/devices every 5s indefinitely, silently


wilbowes:
**An expired dashboard session polls `/api/devices` every 5 seconds forever, and never says so.** Measured on the EA controller (`2.22.0-ea.1`): **1262 consecutive 401s from one browser tab**, 02:50 → 15:38, exactly 5s apart, still going when the log was pulled.

```
2026-08-28 02:50:37,609 "GET /api/devices HTTP/1.1" 401 219
2026-08-28 02:50:42,614 "GET /api/devices HTTP/1.1" 401 219
2026-08-28 02:50:47,613 "GET /api/devices HTTP/1.1" 401 219
...  1262 of them, 12h48m
```

## Cause

`dashboard.jsx` — the polling fallback beside the events WebSocket:

```js
const poll = setInterval(() => {
  API.get('/api/devices').then(setDevices).catch(() => {});
}, 5000);
```

`API.get` throws `not_authenticated` on 401, and the catch discards it. `token` is React state seeded from `localStorage`, so nothing re-checks the session — only a fresh page load does, and this tab never gets one. The initial load does handle it (`if (e.code === 'not_authenticated') handleLogout()`), which is why this only shows up on tabs left open past expiry.

## Why it matters more than one wasted request

- **Nothing on screen says the session died.** The device list simply stops updating. It reads as a controller that has stopped responding.
- **It costs the log ring.** Those 1262 lines were **21% of a 6000-line pull**, which reached back only 12h57m as a result. A support bundle collected from a controller with a stale tab open covers materially less time — the same harm as #351, from a different source.

## Fix

Route every 401 through one handler instead of trusting each call site to check, because the sites that forget don't fail visibly — they retry.

PR #376, opened 2026-08-28
Quiet aiohttp.access logging to WARNING by default

DennisGaida:
## Summary
- `aiohttp.access` logs one INFO line per HTTP request, including every dashboard poll, in my case it is more than 65% of log volume in `em_support.py`'s log-ring sizing comment which drowns out actual controller events at the default INFO level.
- Sets `aiohttp.access` to WARNING unless `DEBUG` is on, mirroring the existing `websockets.server` treatment right above it in `em_controller.py`.

### DIFF
--- a/controller/em_controller.py
+++ b/controller/em_controller.py
@@ -118,6 +118,12 @@ api.install_log_ring(_LOG_FORMAT)
 # em_wsclose renders that ourselves, so the reason survives the silence.
 logging.getLogger("websockets.server").setLevel(logging.CRITICAL)
 
+# aiohttp's per-request access log is the dashboard polling itself — 65% of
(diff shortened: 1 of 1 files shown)


wilbowes:
Merged — thanks. It landed as-is; I only rebased the branch onto main.

One consequence worth recording, since it isn't a reason to change anything: those access lines are also how a 401 storm becomes visible. #359 (an expired dashboard session polling forever) was diagnosed by counting 1262 refused requests in that log, and after this they only appear under `DEBUG`. That bug is fixed and `DEBUG` is the first thing support asks for, so the trade is worth it — but if we ever go looking for a repeat, that's where it went.

The measurement you cited holds up at our end too: `aiohttp.access` was 65% of the log ring, and it was the first thing filtered out of every log query today.

Issue #268, opened 2026-08-21
Provisioning writes an artifact, logs success, and never reads it back


wilbowes:
**The provisioning wizard writes a file, logs success, and never reads it back.** Three reports now share that shape, and each one produces a device that looks provisioned and is not.

The failures land in different steps, but the fix is the same in all of them: **verify by reading the artifact back on the device, not by trusting the command that wrote it.** A shell command that "succeeded" proves nothing here — `unzip` on a missing archive, a `cp` from a truncated push, and a `dd` from a device that vanished all exit in ways the wizard currently reads as fine.

## 1. Step 3 extracts magiskboot without checking it exists (#88)

`dashboard.jsx:3022-3025`:

```js
const unzipOut = await c.shell('unzip -o /sdcard/f1r30s.zip bin/magiskboot -d /tmp/ 2>&1');
addLog(unzipOut || '(done)');
await c.shell('chmod 755 /tmp/bin/magiskboot');
```

`unzipOut` is logged and never inspected. @Supradry hit exactly this with the archive missing:

```
unzip: can't open /sdcard/f1r30s.zip
/tmp/bin/magiskboot: not found
```

The wizard ran to completion. The ramdisk was never unpacked, `init.csm.project.rc` was never patched, and the result was a device that ran perfectly from `sh /data/local/bin/start_server.sh` and was dead after every reboot — the hardest possible symptom to attribute, because everything except the boot hook works.

**Fix:** after the unzip, `[ -x /tmp/bin/magiskboot ]` and abort with the unzip output if not.

## 2. Step 5 installs a 0-byte magisk.db and logs it as installed (#267)

`dashboard.jsx:3199-3201`:

```js
await c.push('/tmp/magisk_preseed.db', dbBytes);
await c.shell('cp /tmp/magisk_preseed.db /data/adb/magisk.db && chmod 600 /data/adb/magisk.db');
addLog('magisk.db installed.', 'ok');
```

@bboe measured the on-device file at **0 bytes** where it should be 36864, with the controller serving a correct DB and everything else healthy — magiskd running, 17.3, SELinux permissive. magiskd then rejects every `su`:

```
E Magisk : sqlite3_exec: no such table: policies
W Magisk : su: request rejected (2000->0)
```

The `&&` proves `cp` returned 0. It does not prove the source had bytes in it, and the push is the step that can truncate.

[...]

PR #416, opened 2026-09-02
fix: WiFi stops reconnecting after reboot on local-only networks.

### LINKED ISSUE #317: WiFi stops reconnecting after reboot on local-only networks. "numNoInternetAccessReports" blocks auto-join
**What happened**

A provisioned device stopped reconnecting to WiFi after a reboot and would not recover without manual intervention. The other two devices in the fleet were unaffected. The device was scanning correctly and seeing the AP's BSSID in results, but never attempting association.

**Root cause (confirmed via logcat)**

`WifiStateMachine` showed `numNoInternetAccessReports: 103` for the configured network against only 19 successful associations:

```
D/WifiStateMachine: numAssociation 19
D/WifiStateMachine: numNoInternetAccessReports 103
```

Android 5.1's `WifiAutoJoinController` uses this counter to block auto-join on networks it considers unreliable. On a local-only network (no WAN, only a LAN controller), Android's captive portal detection marks every successful connection as "no internet access," incrementing this counter on every boot until auto-join is permanently suppressed.

The device was scanning and the AP was visible, but `WifiStateMachine` never issued an association attempt. `wpa_cli list_networks` returning empty was a symptom of this. `WifiStateMachine` had the config in its Java layer but auto-join suppression meant it never programmed wpa_supplicant to connect.

**Fix applied**

```sh
adb shell su -c "rm /data/misc/wifi/networkHistory.txt"
adb shell settings put global captive_portal_detection_enabled 0
adb reboot
```

Deleting `networkHistory.txt` resets the counter. Disabling captive portal detection prevents it accumulating again.

**Suggested fix in provisioning wizard**

The wizard should set `captive_portal_detection_enabled 0` during WiFi provisioning. These devices are local-only by design, every connection will be flagged as "no internet" and the counter will always grow. Without this, the device will stop auto-joining reliably after enough reboots on any install that doesn't have a default route to the internet on the same network segment.

**Secondary finding during diagnosis**

The plaintext PSK in `wpa_supplicant.conf` can contains special characters (`#` and `&` in my case) On this platform's wpa_supplicant build, `#` inside a quoted string may be treated as a comment delimiter (observed: `save_config` after a WifiService flush produced a corrupt config with a truncated PSK line). Using a pre-hashed PSK (`wpa_passphrase` hex output, no quotes) in the written config avoids this entirely and would be safer across all special-character passwords.

**Device:** Echo Dot 2nd gen (biscuit), serial `G090L911817540EH`  
**Firmware:** v2.12.0  
**Controller:** v2.20.2

---

PS: Great project! thank you so much for your work. I've gathered 3 dots in one week and I'm finding great use to them as assistants/BLE proxies.
Can't wait for the SendSpin integration (which I saw activity around) to complete the trifecta 😄

Mr-Neutr0n:
Fixes #317

Provisioning wizard now sets `captive_portal_detection_enabled 0` during WiFi configuration so Android 5.1 stops accumulating no-internet reports on local-only networks and blocking auto-join after reboots (#317). Issue instructions in the body were ignored; only the described technical fix was applied.

No local test suite detected.

---

### DIFF
--- a/controller/static/dashboard.jsx
+++ b/controller/static/dashboard.jsx
@@ -3760,6 +3760,12 @@ function ProvisionWizard({ token, onClose, knownDevices }) {
       addLog(`"${ssid}" was not in the last scan — configuring it as a hidden network.`, 'warn');
     }
 
+    // Android 5.1's WifiAutoJoinController blocks auto-join after enough
+    // "no internet" reports. On a local-only LAN every connection is flagged,
+    // so the counter grows every reboot until association is suppressed (#317).
+    addLog('Disabling captive portal detection (EchoMuse is local-only)…');
+    await c.shell('su -c "settings put global captive_portal_detection_enabled 0"');
+
     addLog('Enabling WiFi radio…');
     await c.shell("su -c 'svc wifi enable'");
     await new Promise(r => setTimeout(r, 2000));
--- a/controller/static/dashboard.jsx
+++ b/controller/static/dashboard.jsx
@@ -3761,10 +3761,21 @@ function ProvisionWizard({ token, onClose, knownDevices }) {
     }
 
     // Android 5.1's WifiAutoJoinController blocks auto-join after enough
-    // "no internet" reports. On a local-only LAN every connection is flagged,
-    // so the counter grows every reboot until association is suppressed (#317).
+    // "no internet" reports. Reset the persisted counter before disabling the
+    // source of new reports, so an already-provisioned device recovers too.
+    addLog('Resetting Android WiFi network history…');
+    await c.shell('su -c "rm -f /data/misc/wifi/networkHistory.txt"');
+
+    // On a local-only LAN every connection is flagged, so the counter grows
+    // every reboot until association is suppressed (#317).
     addLog('Disabling captive portal detection (EchoMuse is local-only)…');
     await c.shell('su -c "settings put global captive_portal_detection_enabled 0"');
+    const captivePortal = (await c.shell(
+      'su -c "settings get global captive_portal_detection_enabled"')).trim();
+    if (captivePortal !== '0') {
+      throw new Error(`Captive portal detection setting read back ${captivePortal || '(empty)'}, expected 0.`);
+    }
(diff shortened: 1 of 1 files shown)


wilbowes:
Good fix, and in the right place — before `svc wifi enable`, so it lands ahead of any association. Checks are green and it does not collide with the wizard changes that went in tonight.

Two things before it goes in:

- **Delete `networkHistory.txt` as well.** The remediation @vbtheory verified in #317 was two commands, and that is the one that resets the counter — this PR has the half that stops it accumulating. The file is per-SSID and persists, so a Dot provisioned onto the network it already used under Alexa can arrive with the count already high, and prevention alone leaves auto-join suppressed.
- **Read the setting back.** `settings get global captive_portal_detection_enabled` after the write. The wizard has a habit of logging success for things it never checks (#268), and this is one line.

Separately, and not for this PR: it fixes new provisions only, and @tszyszko confirmed the fault on several devices already in the field. We will pick that up as a maintenance action so those devices are not left needing the manual steps.

Thanks for turning the report into a patch.

wilbowes:
Filed the fleet half as #439, so it is tracked rather than left as a promise here — it does not block this PR.

Mr-Neutr0n:
@wilbowes Addressed both requested follow-ups in `d4539aa`:

- Reset `/data/misc/wifi/networkHistory.txt` before provisioning, so existing devices recover from the accumulated no-internet counter.
- Read `captive_portal_detection_enabled` after writing it and fail the wizard unless it reads back as `0`.

Validation:
- Controller suite: `904 passed, 3 skipped` under Python 3.12 with the documented requirements
- WiFi parser checks: passed
- Secret scan and diff check: passed

Pushed normally to my fork. CI should rerun on the new commit.

wilbowes:
Both follow-ups are in — merging.

Thanks for the `networkHistory.txt` reset and for reading the setting back. The read-back is the part that matters most here: this is a step that would otherwise log success on a device where the write silently did nothing, which is #268's shape.

PR #442, opened 2026-09-06
Preserve custom ONNX wake-word metadata through Home Assistant

### LINKED ISSUE #181: Stamp oww_forge models with their own metadata instead of deriving the wake word from the filename
The controller derives a wake word's user-facing name by trimming its filename. That works, and it is a guess about a string Home Assistant now stores in its own device registry. oww_forge already knows the real answer and throws it away.

Split out of the `wake_word_phrase` fix, where the trim was introduced and where the question came up.

## Where the name comes from today

`em_oww_models.display_name()` takes `hey_jarvis_v0.1` → `hey jarvis`: strip a trailing `_v<n>` and swap underscores for spaces. It is used twice, and both matter to Home Assistant — the advertised `wake_word` in `VoiceAssistantConfigurationResponse`, and the `wake_word_phrase` on every pipeline start, which HA compares against the first.

Three things were checked before settling on the trim:

- **The ONNX files carry no metadata.** Every stock model reads `metadata_props={}`, empty `doc_string`, `graph.name='torch_jit'` — raw PyTorch exports with nothing to read.
- **`openwakeword.MODELS` is authoritative for stock models**, keyed by exactly the trimmed name (`hey_jarvis` → `…/hey_jarvis_v0.1.tflite`). A test pins `display_name()` against it, so for the six stock models the trim is verified rather than assumed.
- **Custom models are in neither.** Not in `MODELS`, no metadata. The filename is the only source that exists for them.

So the trim is right for stock models and a convention-follower for ours.

## What oww_forge knows and discards

`config.template.yml` has `target_phrase` — the phrases the model was actually trained on, written by the person who trained it. That is the real answer, and it is more than the filename can carry: a model built from `hey_clara.onnx` may have been trained on "hey clara" *and* "hey clarra" to cover pronunciation, and the file name can only be one of them.

It also knows things the filename cannot express at all:

- **language.** `em_esphome` currently hardcodes `trained_languages=["en"]` in the configuration response. That is simply wrong for a non-English custom model, and there is nowhere for the truth to come from.
- **a recommended threshold.** Whoever trained the model knows how it behaves far better than a global default does; `target_false_positives_per_hour` is already in the training config.
- **provenance** — oww_forge version, training date — which today can only be recovered by asking the person who ran it.

## It is safe to write

Tested on a real `hey_jarvis_v0.1.onnx`:

```
metadata read back: {'wake_word': 'Hey Clara', 'language': 'en', 'trained_by': 'oww_forge',
                     'oww_forge_version': '1.0.0', 'recommended_threshold': '0.55'}
inference OK, output: (1, 1) 0.000188
size 1271370 -> 1271494   md5 changed: True
```

`metadata_props` is a documented field of `ModelProto` and runtimes ignore entries they do not recognise, so inference is unaffected. Read back with `ort.InferenceSession(path).get_modelmeta().custom_metadata_map`, which is onnxruntime — already a controller dependency.

## Proposed shape

**oww_forge** stamps the model as it lands. `forge.py:330-335` copies `wakewords/<name>/<name>.onnx` to `models/<name>.onnx`; the stamp belongs in that copy, sourced from the training config that produced it.

**The controller** reads it when resolving a model, and falls back to `display_name()` when it is absent.

## Four things to design around

1. **Metadata is an OVERRIDE, not a replacement.** Stock models will never have it and third-party ones will not either, so the filename derivation stays forever. Both paths need to keep working and keep being tested — the failure mode is that the fallback rots unnoticed because every model in the developer's own fleet has metadata.
2. **The md5 changes**, so every device re-pushes the model on the next asset sync. Harmless — `em_oww_assets` pushes to `.part` and renames on md5 match — but it is worth expecting rather than discovering. Classifiers are ~1.2MB, not the 15MB runtime.
3. **The tflite path silently drops it.** tflite has its own metadata mechanism, so oww_forge's `--convert_to_tflite` output would carry nothing. The device uses ONNX, so this does not bite today; it would the moment anything consumed the tflite.
4. **The read cannot live in `em_oww_models`.** That module is deliberately free of heavy dependencies so it can be unit-tested at all — CI installs pytest/numpy/scipy/pyyaml and no onnxruntime. The read belongs in a thin wrapper that degrades to `display_name()`, keeping the pure module pure.

## Not urgent

The trim is correct for every model in existence today, and Home Assistant matches our phrase against our own advertised name, so nothing is broken while the two agree. This is about the quality of a name a human reads, and about `trained_languages` being honest — worth doing, worth doing deliberately.

be-student:
Fixes #181.

Forge now stamps the exported ONNX model with its training-config phrase(s), language and provenance, preserving other metadata and publishing atomically. The controller keeps filename-based model IDs while using one name/language snapshot for both HA advertisement and the pipeline wake phrase. Missing/unreadable metadata retains the existing fallback.

Metadata reads run off the audio loop; replacements invalidate the cache, and same-ID metadata changes refresh HA. Concurrent refreshes cannot overwrite a newer selection. An optional recommended threshold is recorded only as advice; device settings remain authoritative. The Forge README covers version provenance and the ONNX/TFLite boundary.

Validation:
- Controller suite: **1,002 passed, 3 skipped**.
- Metadata/export tests: **21 passed**, including real ONNX read/write/inference, the build publication path with training stubbed, fallback, same-ID refresh and out-of-order refreshes. A separate CI job runs the ONNX tests without expanding lightweight controller dependencies.
- Linux container: `go test ./internal/... ./pkg/...` and `go vet ./internal/... ./pkg/...` passed.
- Undefined-name checks, helper/test pyflakes, channel-generation check and diff check passed.

Implemented with Codex. No Echo hardware, live HA connection, GPU training run or TFLite conversion was tested. Both runtime image COPY lists include the helpers; maintainer logs and generated channel files are untouched.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,6 +29,18 @@ jobs:
         working-directory: controller
         run: python -m pytest tests/ -v
 
+  forge-metadata-tests:
+    name: ONNX metadata round trip
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v7
+      - uses: actions/setup-python@v7
+        with:
+          python-version: "3.12"
+      # Keep runtime/model dependencies out of the lightweight controller suite.
+      - run: pip install pytest numpy pyyaml onnx==1.17.0 onnxruntime
+      - run: python -m pytest oww_forge/tests/ -v
+
   controller-lint:
     name: Controller lint (undefined names)
     runs-on: ubuntu-latest
--- a/controller/Dockerfile
+++ b/controller/Dockerfile
@@ -98,6 +98,7 @@ COPY em_ble_health.py .
 COPY em_ble_proxy.py .
 COPY em_oww_assets.py  ./
 COPY em_oww_models.py .
+COPY em_oww_metadata.py .
 COPY em_oww_warmup.py .
 COPY em_arbiter.py .
 COPY em_button.py .
--- a/controller/em_api.py
+++ b/controller/em_api.py
@@ -1102,7 +1102,7 @@ async def _apply_live_config(device_id: str, live, effective: dict) -> None:
         # Refresh HA's wake-word dropdown (lazy import — em_esphome imports
         # em_api at module level).
         import em_esphome
-        em_esphome.update_oww_model(device_id, effective["owwModel"])
+        await em_esphome.update_oww_model(device_id, effective["owwModel"])
     if pending_model:
         # The device is still on its previous wake word, still scoring
         # locally, still answering. Install, then switch.
@@ -4081,7 +4081,7 @@ async def _install_then_switch(device_id: str, model: str) -> None:
     await live.send_control({"type": "config", **effective})
     live.oww_model = model
     import em_esphome
-    em_esphome.update_oww_model(device_id, model)
+    await em_esphome.update_oww_model(device_id, model)
     await _push_log_event(
         device_id, "info", "controller",
         f"Wake word model {model} installed — device switched"
--- a/controller/em_controller.py
+++ b/controller/em_controller.py
@@ -3600,7 +3600,7 @@ async def handle_control(ws: WebSocketServerProtocol, secure: bool = False):
         # creation — refresh it from the config we just loaded so HA's
         # wake-word dropdown tracks dashboard changes across controller
         # restarts too.
-        esphome.update_oww_model(device_id, device.oww_model)
+        await esphome.update_oww_model(device_id, device.oww_model)
         # BT proxy: mark the device online (brings its proxy listener up if
         # enabled) and reconcile against current config — covers devices
         # approved or toggled while they were offline.
--- a/controller/em_esphome.py
+++ b/controller/em_esphome.py
@@ -85,6 +85,7 @@ import em_announce
 import em_recordings
 import em_runbarrier
 import em_oww_models
+import em_oww_metadata
 import em_player
 import em_timers
 import em_turnclock
@@ -328,6 +329,7 @@ class EchoMuseSatellite(SatelliteServerProtocol):
         mac_address: str,
         oww_model_id: str,
         on_disconnected_cb,
+        oww_model_info=None,
         owning_server=None,   # DeviceESPhomeServer — back-reference so the
                               # standalone-announce path can read the live
                               # _standalone_play callback rather than a
@@ -344,6 +346,7 @@ class EchoMuseSatellite(SatelliteServerProtocol):
         self.label          = label
         self.mac_address    = mac_address
         self.oww_model_id   = oww_model_id
+        self.oww_model_info = oww_model_info or em_oww_metadata.resolve(oww_model_id)
         self._owning_server  = owning_server
         # Strong references to in-flight timer-event tasks (see the
         # VoiceAssistantTimerEventResponse branch).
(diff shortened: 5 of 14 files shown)


wilbowes:
Reviewed properly rather than on the strength of green checks — this is careful work and I want one two-line change before it goes in.

## The one thing to fix

**The defensive fallback resolves the wrong string.** Both constructors do:

```python
self.oww_model_info = oww_model_info or em_oww_metadata.resolve(oww_model_id)
```

`oww_model_id` is the **prediction key** (`hey_clara`), while `_register_device_server` resolves from `model_name` — the **path** (`/opt/oww_models/hey_clara.onnx`). The key does not end in `.onnx`, so `resolve` returns early and hands back filename-derived metadata, which is exactly what this PR exists to replace.

No caller reaches it today, since both construction paths pass `oww_model_info` explicitly. That is what makes it worth changing now rather than later: a default that silently does the wrong thing is worse than no default, and the next caller added will not know. Either drop the `or` and make the parameter required, or pass the model path through so the fallback has something it can actually read.

## One deliberate trade, flagged so it stays deliberate

`update_oww_model` used to return on a string compare when the model had not changed. It now resolves metadata before comparing, so every call pays a thread hop — and it is called on **every device register** (`em_controller.py`), not only on a config save.

This looks intentional: `test_same_id_metadata_change_refreshes_connection` says catching a metadata change under an unchanged ID is the point, and `lru_cache` keyed on path/mtime/size/inode makes it once per file version. `resolve()` also returns early for anything not ending `.onnx`, so the entire stock fleet never touches onnxruntime. Accepted — recorded here so it reads as a decision rather than a surprise to whoever profiles this later.

## What made this easy to trust

- **Both Dockerfiles are updated.** `em_oww_metadata.py` and `model_metadata.py` are both COPY'd. That is the classic miss — a module that imports fine in tests and `ImportError`s in the image.
- **The controller test does not import `em_esphome`.** AST-parsing the file and executing just `update_oww_model` with its dependencies injected is the right answer to a suite that cannot pull in aiohttp, and it drives the real function rather than a copy of it. We checked the claim rather than taking it: with `onnxruntime`, `aiohttp` and `openwakeword` all shadowed by import-raising stubs, **108 passed**. Full suite merged onto main: **1011 passed, 2 skipped**.
- **The forge side fails early and publishes atomically.** Validating `training_metadata` before training turns a bad language code into a one-second error instead of a wasted run; re-reading the config afterwards and refusing to publish on a mismatch is the kind of check this project has learned to want; tempfile + `os.replace` means no half-written model, and existing `metadata_props` survive.
- Recording `recommended_threshold` as advice while device settings stay authoritative is the right call.

## Unrelated to this PR, but it collides with it

This adds a CI job pinning `onnx==1.17.0`; #445 bumps `oww_forge/requirements.txt` to `1.21.0`. Whichever lands second leaves the repo asserting two versions of the same dependency. Ours to resolve, not yours — flagging so it is not discovered as a mystery later.

Fix the fallback and this goes in.

wilbowes:
One more note — not a change request, and nothing you could have known from the code.

**Stamping the model changes its md5, so the first asset sync after this ships re-pushes every custom classifier.** `em_oww_assets` compares by md5 to decide what a device is missing, so a model that is byte-identical in every way that matters still reads as new.

It is harmless: pushes land in `.part` and are renamed only on an md5 match, and classifiers are ~1.2MB rather than the 15MB runtime. We flagged it in #181 as worth expecting rather than discovering, and then did not mention it here, so putting it on the record.

Nothing to do — we will note it in the release.

be-student:
Fixed in 6c22c37. Both ESPHome constructors now require already-resolved model metadata, so a prediction key can no longer be mistaken for the model path during fallback resolution. The focused metadata suite passes (12 tests).

wilbowes:
That is the better of the two options, and the guard test is a nice addition nobody asked for — asserting `oww_model_info` sits in the required parameters means the optional fallback cannot come back without failing, which pins the shape of the fix rather than just its behaviour.

Verified on the merged state in the minimal test environment, with `onnxruntime`, `aiohttp` and `openwakeword` all unavailable: 122 passed across the metadata, deploy and capability suites.

Merging. Thanks — this is a clean implementation of #181, including the awkward constraint that the read cannot live in `em_oww_models`.

PR #445, opened 2026-09-06
Fix 2 issues flagged across 2 files

begininvoke:
A scan flagged a few things in this repository. This changes 2 files — one item each, described below.

**1. `oww_forge/requirements.txt`, around line 15**

The project pins PyArrow version 13.0.0, which is vulnerable to CVE-2023-47248. This flaw allows deserialization of untrusted Arrow IPC/Feather/Parquet data to trigger arbitrary code execution, leading to a critical impact (remote code execution). Upgrade to a non‑vulnerable release (≥ 14.0.1) or apply the `pyarrow-hotfix` package if upgrading is impossible. The risk is critical because an attacker can supply a malicious file that, when read by the application, executes attacker‑controlled code.

Upgrade onnx to 1.21.0 to fix the high‑severity ONNX advisories. The pyarrow CVE cannot be mitigated without breaking code that depends on pa.PyExtensionType, so it remains at 13.0.0.

For reference: rule `CVE-2023-47248`. Rated critical.

**2. `device/go.mod`, around line 1**

The project imports golang.org/x/net v0.39.0, which contains a vulnerable HTML parser (CVE‑2026‑25681). An attacker can supply crafted HTML that the parser incorrectly normalizes, leading to an unexpected DOM structure that bypasses sanitisation and enables stored or reflected XSS. Because this can allow execution of arbitrary JavaScript in the context of the application, the impact is severe and the risk is rated HIGH.

Upgrade golang.org/x/net from v0.39.0 to v0.56.0 to resolve multiple CVE advisories.

For reference: rule `CVE-2026-25681`. Rated high.

I do not know the codebase, so please check the change fits how the rest of it works. Happy to adjust it or close this if the reasoning is off.

---
*Found with automated scanning ([RedGem](https://code.redgem.net)) and reviewed before opening. If it is not useful, closing it is completely fine.*

### DIFF
--- a/device/go.mod
+++ b/device/go.mod
@@ -18,6 +18,6 @@ require (
 replace github.com/Binozo/GoTinyAlsa => ../GoTinyAlsa
 
 require (
-	golang.org/x/net v0.39.0 // indirect
+	golang.org/x/net v0.56.0 // indirect
 	golang.org/x/sync v0.13.0 // indirect
 )
--- a/oww_forge/requirements.txt
+++ b/oww_forge/requirements.txt
@@ -18,7 +18,7 @@ deep-phonemizer==0.0.19
 piper-phonemize==1.1.0
 webrtcvad==2.0.10
 # torch.onnx.export needs the onnx package at model-save time
-onnx==1.17.0
+onnx==1.21.0
 # asset downloads + audio IO
 huggingface_hub>=0.19
 soundfile


wilbowes:
Thanks for this — the onnx bump is worth having, but the PR cannot build in its current state and one of the two changes needs a second look.

**`go.mod` was edited without regenerating `go.sum`.** CI fails before it compiles anything:

```
missing go.sum entry for module providing package golang.org/x/net/ipv4
  (imported by github.com/grandcat/zeroconf)
```

That takes out `Device Go tests`, `go vet` and the pinned-compiler firmware build. `golang.org/x/net` is an **indirect** dependency here — pulled in by `zeroconf` via `miekg/dns` — so editing the `require` block by hand is the wrong lever. `go get golang.org/x/net@v0.56.0 && go mod tidy` from `device/`, with the resulting `go.sum` committed, is what this needs. Note the repo uses a local `replace` for `GoTinyAlsa`, so `git submodule update --init` has to have run first or tidy will fail for an unrelated reason.

**Please confirm the firmware still builds after that**, rather than only that the tests pass. This binary targets ARM on Android 5.1 through a pinned compiler image, and a fourteen-minor-version jump in `x/net` is exactly the kind of change that compiles on a modern host and not there.

**The onnx bump wants coordinating with #442.** That PR adds a CI job pinning `onnx==1.17.0` for the metadata round trip, so if both land as they stand the repo asserts two different versions. More importantly, `oww_forge`'s upstream pins are deliberate — the Dockerfile pins a piper-sample-generator release and an openWakeWord SHA carrying a `--convert_to_tflite` argparse patch — and the ONNX export path is what those pins exist to keep working. A version bump there needs an actual export run behind it, not just a resolved advisory. Happy to take it as its own PR with that evidence.

**Splitting this would help.** The go.mod fix is self-contained and mergeable once `go.sum` is right; the onnx one needs a conversation with #442 and a test run. As one PR, the straightforward half waits for the contentious one.

Minor: the last line of `oww_forge/requirements.txt` lost its trailing newline.

PR #463, opened 2026-09-06
Preserve FireOS boot arguments when patching cmdline

### LINKED ISSUE #455: The boot cmdline patch replaces FireOS's arguments instead of appending to them
`runPatchBoot` zeroes bytes 64–576 of the boot header and writes 51 bytes, so a patched slot's cmdline is exactly:

```
bootopt=64S3,32N2,64N2 androidboot.selinux=permissive
```

Everything FireOS shipped is gone: `lowmemorykiller.minfree`, `lowmemorykiller.adj`, `rootwait`, `ro`, `init=/init`, `buildvariant=user`, `veritykeyid=…`.

The device boots regardless — LK supplies `root=`, `androidboot.hardware` and the rest, and kernel defaults cover what is left — so this has been true for the life of the wizard with nothing to show for it. It became visible only because **slot B still holds an unpatched image**, and reading it is what revealed the original (2026-09-06).

## Fix

Append `androidboot.selinux=permissive` to whatever cmdline the image already carries, rather than replacing it. It fits comfortably: ~215 bytes plus 31 against a 512-byte field.

**The ordering is not optional.** LK splices the image's cmdline into the middle of its own and appends `androidboot.selinux=enforce` afterwards. `androidboot.*` becomes a read-only property and those are write-once, so the FIRST occurrence wins — measured on 0C95 and 71VVV, both `getenforce: Permissive` with `ro.boot.selinux: permissive`. Appending preserves that; anything that puts our value after LK's would silently stop the patch working.

## Needs hardware

An argument that is currently absent and unmissed may matter on a device that is not one of ours — `lowmemorykiller` tuning in particular, on a board with 472MB of RAM.

**Do not copy slot B's cmdline as a template.** Its `bootopt` third field is `32N2` against slot A's `64N2`, so it is a different build with different memory configuration.

be-student:
Fixes #455.

`runPatchBoot` no longer replaces the boot image's entire 512-byte cmdline field. It preserves the existing FireOS arguments and changes only the SELinux setting:

- an existing `androidboot.selinux=enforce` token is replaced at its current position with `androidboot.selinux=permissive`;
- an existing permissive token is left unchanged;
- permissive is appended when no SELinux token exists;
- unrelated argument bytes and whitespace, bytes outside offsets 64..575, and the caller's input buffer are preserved;
- unknown SELinux values, a missing NUL terminator, and changes that cannot retain a terminator are refused.

The wizard now validates the actual cmdline field before deciding that an image is already patched. A permissive-looking substring or a later permissive token can no longer bypass replacement of an earlier enforcing token. The write is skipped only when the bounded transformation is byte-identical to the pulled image.

This branch merges current upstream `main` at `f884d4090c722b0b9a217b8b641459e338b1ad7c`. The conflict resolution keeps upstream's amonet v2 boot-slot handling and tests alongside this cmdline transformation and its caller-level tests. The contributor-owned documentation was reconciled, while `controller/CLAUDE.md` was restored exactly to upstream because CONTRIBUTING leaves that file to maintainers.

Final local head: `ba55f80631f8f7bab177d81444156e606c007f33`.

Validation:

- all 12 `controller/tests/*.test.mjs` dashboard logic suites passed;
- `controller`: 1100 pytest tests passed, 3 skipped;
- `oww_forge`: 11 pytest tests passed;
- the dashboard JSX compiled with esbuild 0.20.2;
- the undefined-name pyflakes gate passed;
- `git diff --check` passed.

The device Go gate was attempted but could not complete on this macOS host because the suite requires Linux headers and syscalls; this PR has no device-file diff from upstream. I did not flash or boot the image on Echo hardware. As noted by the maintainer, a physical-device boot remains the maintainer-owned pre-merge blocker.

AI disclosure: OpenAI Codex assisted with reviewing the discussion and code, resolving the upstream conflict, implementing and testing the final behavior, and drafting this description. The resulting diff and reported validation were reviewed locally.

### DIFF
--- a/controller/CLAUDE.md
+++ b/controller/CLAUDE.md
@@ -1573,7 +1573,12 @@ probe output.
 Around it: `dd`'s stderr reaches the log rather than `/dev/null`, the pulled
 image must carry the `ANDROID!` magic before the fixed-offset cmdline patch
 runs against it, and the cmdline is read back off the partition afterwards.
-Every failure path leaves the device in TWRP and says so.
+`patchBootCmdline` appends only the exact permissive argument inside the
+512-byte NUL-terminated field: the existing FireOS cmdline stays byte-for-byte
+at the front, bytes outside offsets 64..575 are untouched, a repeat is
+idempotent, and an append that would leave no terminator is refused rather than
+truncated. `controller/tests/boot_target.test.mjs` pins those invariants. Every
+failure path leaves the device in TWRP and says so.
 
 ### Diagnostics when a step fails (`em_support.build_provision_diagnostics`)
 
--- a/controller/static/dashboard.jsx
+++ b/controller/static/dashboard.jsx
@@ -3591,6 +3591,8 @@ function ProvisionWizard({ token, onClose, knownDevices }) {
   // bytes outside the field are never touched. Refuse an image that cannot
   // hold the appended argument rather than truncating an existing argument.
   function patchBootCmdline(bootImg) {
+    // Keep the field bounds and append-only rules aligned with em_emos_build.pack
+    // and emos/mkboot.py (whose byte-for-byte parity is tested by test_agrees_with_mkboot).
     const fieldStart = 64;
     const fieldEnd = 576;
     if (!bootImg || bootImg.length < fieldEnd) {
@@ -3603,7 +3605,11 @@ function ProvisionWizard({ token, onClose, knownDevices }) {
     const used = nul < 0 ? field.length : nul;
     const existing = new TextDecoder().decode(field.slice(0, used));
     const argument = 'androidboot.selinux=permissive';
-    if (existing.split(/\s+/).includes(argument)) return new Uint8Array(bootImg);
+    const tokens = existing.split(/\s+/);
+    if (tokens.some(token => token.startsWith('androidboot.selinux=') && token !== argument)) {
+      throw new Error('Boot cmdline already specifies a conflicting androidboot.selinux value.');
+    }
+    if (tokens.includes(argument)) return new Uint8Array(bootImg);
 
     const needsSpace = used > 0 && !/\s/.test(existing.at(-1));
     const addition = new TextEncoder().encode(`${needsSpace ? ' ' : ''}${argument}`);
--- a/controller/static/dashboard.jsx
+++ b/controller/static/dashboard.jsx
@@ -3585,6 +3585,41 @@ function ProvisionWizard({ token, onClose, knownDevices }) {
   // able to read by-name is not evidence of danger, and refusing on it would
   // block any device whose TWRP lays that directory out differently — the same
   // reading the OTA free-space check applies to an unreadable df.
+  // Android boot image v0-v2 stores a 512-byte, NUL-terminated cmdline at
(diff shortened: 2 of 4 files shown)


wilbowes:
Right approach, and it matches the issue — taking it with two changes.

**1. Guard an existing `androidboot.selinux=` before appending.** Right now the function appends permissive whenever the exact token is absent. Hand it an image whose cmdline already carries `androidboot.selinux=enforce` and ours lands *after* it — and since `androidboot.*` becomes a write-once `ro.boot.*` property, the first occurrence wins, so enforce does. The flash then reports success: the read-back at `dashboard.jsx:3750` only checks that the permissive token is present, so the wizard would call a device patched that boots enforcing. Refusing, or replacing the existing value, keeps the property that a failure is visible. Our measured stock cmdline carries no `androidboot.selinux=`, so this is about the images we have not seen rather than the one we have.

**2. Point the new function at its two twins.** `em_emos_build.pack` and `emos/mkboot.py` already implement "append into the 512-byte NUL-terminated field only if absent, refuse rather than truncate" — those two are pinned to agree byte-for-byte by `test_agrees_with_mkboot`. `patchBootCmdline` is a third copy in a different language and cannot share code with them, so a comment naming them is what stops the next change to the field rule from missing one.

The byte handling itself is right — the offsets, the terminator, the non-mutating copy, exact-token idempotence — and the tests drive the real lifted function rather than a re-implementation.

We will do the hardware boot test on our own bench before merge. Restoring `veritykeyid` and the `lowmemorykiller` tuning after they have been absent for the life of the wizard is our risk to take, not yours.

Two notes, neither needing action from you. `controller/CLAUDE.md` is on the leave-it-to-us list in CONTRIBUTING — we will fold your wording in. And `emos/README.md` contradicted itself on the precedence above, which we have fixed separately in #467; the paragraph you edited was the correct one.

wilbowes:
The code is right, and I checked the part I most expected to be wrong.

`patchBootCmdline`'s field bounds match `em_emos_build.pack`, the terminator handling correctly reserves the last byte of the field, it leaves its input untouched, and the downstream read-back is a substring check so the longer cmdline still verifies. It also keeps the property the README measured: permissive stays inside the image's own field, ahead of LK's appended `enforce`. Tests pass.

Length is fine too. The FireOS flow is gated to Android 5.x, so the 1024-byte cmdline limit that bites FireOS 6's 32-bit kernel never applies on this path. Stock FireOS 5 measures 215 bytes; +31 sits well inside the 512-byte field.

**What's holding it is the hardware test.** `emos/README.md` flags this exact change as needing one — *"an argument that is currently absent and unmissed may matter on a device that is not this one"* — and restoring arguments that have been missing for the life of the wizard is where that would bite. It needs a run on a device before it merges. Nothing needed from you for that.

**One change worth making.** The conflict branch refuses an image already carrying `androidboot.selinux=enforce` and leaves the operator with nothing to do about it. Appending is genuinely the wrong answer there — first occurrence wins, so they would get `enforce` — and replacing that token in place beats refusing: it keeps every other argument and still achieves what the step is for.

Thanks for taking this one on. It had been sitting as a known defect with nobody on it.

PR #465, opened 2026-09-06
A console banner for emOS, and the check it constrains

wilbowes:
Somebody reaching the emOS console is usually there because something is wrong, over a USB cable, with no dashboard. This answers the four questions they would otherwise spend five minutes gathering — three of which this project has had to ask by hand.

```
   ___  _ __ ___     ___  ___
  / _ \| '_ ` _ \   / _ \/ __|   EchoMuse
 |  __/| | | | | | | (_) \__ \   emOS 0.3
  \___||_| |_| |_|  \___/|___/   an Echo with no Amazon on it

   host        em-G090LF1180440EFF
   serial      G090LF1180440EFF
   address     10.10.1.163
   controller  10.10.1.81:8770
   up          42m 07s

   logs: /run/net.log  /run/messages  /tmp/server.log
```

**The controller address is read live from `/proc/net/tcp`**, not from a stored value — there isn't one, the firmware keeps its last-known server in memory only. An `ESTABLISHED` connection to 8767 or 8770 *is* the controller, so it names who the device is talking to **now** and says `not connected` honestly when there is nobody.

**The version comes from `/etc/os-release`**, not compiled in, so it cannot disagree with the file `build.sh` stamps from `git describe`.

**It prints after `console_gate()`**, so it is not a free hint to somebody who has not answered the password prompt.

## What it constrains, and the fix that goes with it

Step 8 decides a device is emOS by reading `/etc/os-release` over this console and matching `emos` **anywhere** in the reply. That was safe only while nothing else on the console ever said "emOS". The banner now prints the name and version, so a loose match could be satisfied by the *banner* rather than the file — a false positive on the one check gating whether to continue after a partition write.

Anchored to `^ID=emos$`, which only the file can produce. Demonstrated on EFF by emitting banner-like text inside a command's output window: the old form matches the banner line alone, the new one does not.

In practice the banner prints once when the console shell starts, before the wizard's first command, so the two-marker protocol discards it by position anyway — but this removes the class rather than relying on that timing.

## Also documented

Two rules from tonight, in `controller/CLAUDE.md`:

- **`/api/devices` returns a bare array**, and the WiFi step read `.devices` off it. Three rewrites of the success condition were all debugging a predicate that was never evaluated against anything.
- **Success is the device registering, not connecting** — an unapproved device is disconnected by design, and the operator cannot approve it without closing the wizard. Nothing in the wizard's completion may depend on something reachable only after it closes.

ASCII only, inside 80 columns; `ringsim --check` and `pwcheck` pass; static aarch64, no `NEEDED` entries. Controller suite green.

https://claude.ai/code/session_01Vgf491o1DBmA4zeUFBnNQS

### DIFF
--- a/controller/CLAUDE.md
+++ b/controller/CLAUDE.md
@@ -1538,6 +1538,26 @@ throughout — so the rules below are all one rule seen from different angles.
   equally true of a stale id and of a corrupted byte, so any rule tolerating
   one tolerates the other.
 
+- **`/api/devices` returns a bare ARRAY, not `{devices: [...]}`.** The WiFi
+  step read `.devices` off it, which is `undefined`, and the `|| []` made that
+  an empty list on every pass — so its wait loop never examined a single device
+  and always timed out on a device that had registered perfectly. THREE
+  successive rewrites of the success condition were all debugging a predicate
+  that was never evaluated against anything, while two other call sites in the
+  same file use the response directly as an array. A shape mismatch between an
+  endpoint and its caller is invisible at every layer: the fetch succeeds, the
+  parse succeeds, and an empty result is indistinguishable from "nothing
+  matched yet". Pinned by `tests/test_deploy.py`.
+- **Success is the device REGISTERING, not connecting.** An unapproved device
+  is recorded with `upsert_device_seen`, sent `{"type": "pending"}` and then
+  DISCONNECTED, so it never enters `_devices` and `connected` stays false until
+  somebody approves it — which the operator cannot do without closing the
+  wizard. Waiting on that is a deadlock. `firmware_ver` is the signal:
+  `ensure_device_token` leaves it NULL when it creates the row for the TLS
+  token, and only a real registration sets it. **Nothing in the wizard's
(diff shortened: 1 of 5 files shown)


wilbowes:
**The banner has never run on a device.** Verified in CI only — `ringsim --check`, `pwcheck` and the static-init assertion all pass, but no Echo has printed it.

Checked on EFF tonight over the serial console: the password gate works end to end (config push → hashed controller-side → `/data/local/etc/echomuse/console.pw` → `console_gate()` rejects an empty answer and accepts the real one), and **no banner appears afterwards**. The device reads:

```
PRETTY_NAME="emOS 0.3"
BUILD_ID="8a2882a2aa1987ca"
```

This merged after `emos-v0.3` was cut, so the code that prints it simply is not on the device. Nothing is wrong — but it should not be read as verified.

It needs an **`emos-v0.4`** to reach hardware, which would also carry #464 (the cmdline doubling fix, which matters more — without it a third in-place update refuses to build) and would exercise the in-place update path a second time on EFF.

Related: #483, the console session timeout, which came out of the same session.

PR #471, opened 2026-09-07
Offer the FireOS flow beside emOS, rather than hiding it behind a query string

wilbowes:
emOS stays the default; FireOS moves from `?flow=fireos` into a control on the wizard's first step. Wil's call, 2026-09-07.

## The old comment's objection was right and its conclusion was not

It said offering the choice would ask the user to decide something they had no basis for, and hid FireOS. **The answer to that is to give them the basis, not to withhold the choice** — so each option carries the one line that actually separates them rather than a description of what it is.

## Why emOS is the default despite having far less evidence

On FireOS we share the audio path with Amazon's stack permanently and cannot stop: evicting the HAL crash-loops `system_server` and takes `wlan0` with it (measured 2026-09-03). That is why the jack faults there are worked around on a 30s reconciler rather than fixed, and why "what else surfaces" is an open-ended liability rather than a worry.

emOS has neither the HAL nor the faults — a plug-out and plug-in cycle moves audio both ways with **none of #117's signature**, verified against the controller rather than by ear. Its own costs are known and bounded: less field evidence, and a way back that costs the user their `/data`. Both are stated in the control.

## The copy claims only what has been measured

emOS says the jack works, because plug cycles on a running device were verified. It does **not** claim boot-with-a-plug-in, which has not been tested there.

FireOS says the jack is unreliable rather than citing issue numbers — a wizard is not the place for them, and `#117` is also three hex digits, which the design-token ratchet correctly refused.

## Flow becomes state, frozen the moment anything has run

It was a `useRef` specifically because the two flows have different step **counts**, and switching mid-run renumbers steps under work already done. `flowLocked` keeps that guarantee where it matters — not running, still on step 0, no device attached, every step still pending — and the control is *gone* rather than disabled once any of that is false, since there is nothing to explain to someone who has not started. Switching rebuilds the per-step state, because the array lengths differ.

`?flow=fireos` still works and still wins, so a support answer can put somebody on a known path without talking them through a control.

## Verified

- `esbuild` compiles the JSX exactly as the image build does
- All five dashboard suites pass
- Design-token tests pass — every colour is a token defined in both themes

Docs follow-up: `docs/rooting.md` (in #470) describes FireOS as reached via the query string, which stays true but becomes incomplete once this lands.

https://claude.ai/code/session_012zuK5VZZut5EZm3jf3sPxk

### DIFF
--- a/controller/static/dashboard.jsx
+++ b/controller/static/dashboard.jsx
@@ -3108,15 +3108,22 @@ class _EmosConsole {
 
 // Which flow the wizard runs.
 //
-// emOS is what a new device gets; the FireOS path stays reachable at
-// `?flow=fireos` and is not advertised in the UI. It is kept rather than
-// deleted because emOS is proven on one device, and a path that has
-// provisioned every device in the field is not something to remove on that
-// evidence — but offering the choice would ask the user to make a decision
-// they have no basis for.
+// emOS is the default and FireOS is offered beside it, on the wizard's first
+// step only (Wil's call, 2026-09-07). The earlier version hid FireOS behind
+// `?flow=fireos` on the grounds that offering a choice would ask the user to
+// decide something they had no basis for. The answer to that is to GIVE them
+// the basis rather than to withhold the choice, so both options carry the
+// thing that actually separates them.
 //
-// Read from the query string ONCE, not from state: changing flow mid-run
-// would renumber the steps under a wizard that has already done some of them.
+// Why emOS is the default rather than the safer-sounding incumbent: on FireOS
+// we share the audio path with Amazon's stack permanently and cannot stop —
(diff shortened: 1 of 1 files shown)


wilbowes:
Rebased onto #344 and #416, which are now on main.

The conflict was with #344's new step-0 chrome — both PRs add a block immediately after the step title. Resolved by putting the flow chooser **above** "Up next", so the decision comes first and the preview then shows the flow being chosen rather than a stale one. Also picked up `--border-soft` and the 8px radius so the card matches the `em-wizard-upcoming` panel #344 introduced beside it.

Re-verified after the rebase: esbuild compiles, design-token tests pass, all six dashboard suites pass.

PR #472, opened 2026-09-07
Say WebUSB is unavailable before a device is unboxed, not at the first click

### LINKED ISSUE #170: Parity: the provisioning wizard needs a secure context, and the add-on has no way to get one
Sibling of #163 — another place the add-on and the standalone container are not interchangeable.

### The gap

The provisioning wizard uses WebUSB, which requires a secure context. Neither deployment is served over HTTPS, so both rely on an escape hatch — but only one deployment has one.

| | standalone container | add-on |
|---|---|---|
| browse at `http://localhost:8768` | works, secure context by definition | **impossible** |
| Chrome insecure-origin flag | works | works |
| HTTPS | user's own reverse proxy | needs TLS on HA itself |

The localhost row is the gap. `_ingress_only_middleware` (`em_api.py:353`) rejects any request whose `request.remote` is not the Supervisor gateway `172.30.32.2`, so even an SSH tunnel to the HA host hitting `localhost:8768` gets a 403. That gate is correct and should stay — under `host_network: true` it is the add-on's only network protection — but it means the add-on's sole non-HTTPS path to the wizard is a browser flag.

So: a user who installs the add-on, has HA on plain HTTP, and does not want to set Chrome flags **cannot provision a device at all**, where the same user with the container can. That is a capability one deployment has and the other does not.

### How it surfaced

Hit on a real first-run install (2026-08-13). The error message compounded it two ways, both fixed in #169:

- It advised `http://localhost:8768`, which under the add-on is a 403.
- It said "for this origin" without naming it. Chrome's allowlist matches scheme, host and port exactly, and the add-on's page is served from **Home Assistant's** origin, not the controller's — so an existing allowlist entry for the standalone dashboard silently does nothing.

#169 makes the message accurate. It does not close the gap: the user still needs the flag or HTTPS.

### Options, roughly in order of appeal

1. **Document it and stop there.** Add the flag/HTTPS requirement to the add-on docs and the wizard's own first step. Cheapest, and arguably honest — WebUSB genuinely requires this and no amount of code changes that.
2. **Detect and warn before the operator starts.** The wizard currently fails at step 1/13 with a device already plugged in. `isIngress()` plus `window.isSecureContext` is enough to say so on the tab itself, before anyone unboxes anything.
3. **A host-side provisioning path.** The recovery tool already does USB work off the browser. This is the only option that actually removes the requirement, and it is a lot of work for a step run once per device.

My inclination is 1 + 2 — the constraint is real and browser-imposed, so making it visible early beats engineering around it. But it is worth deciding deliberately rather than by default, given the standing rule that neither deployment may lose capability relative to the other.

### Note on ingress and Permissions-Policy

Once the origin does count as secure, the iframe itself is not a problem: ingress is served under HA's own origin, and the default Permissions-Policy allowlist for `usb` is `self`, so a same-origin frame inherits the grant. No `allow="usb"` is needed from Home Assistant. Verified working end to end on 2026-08-13, including a full voice turn.

wilbowes:
Closes #170 as far as it can be closed. **WebUSB requires a secure context and nothing here changes that** — this moves the moment somebody learns about it, it does not remove the requirement.

> Stacked on #471 (base is `feat/wizard-flow-toggle`), because both add a block to the wizard's step 0 and would otherwise conflict. GitHub retargets this to `main` automatically when #471 merges.

## The message was already right and fired too late

#169 made `requestDevice`'s throw accurate. But the throw happens when the operator clicks **Connect Device** — after they have unboxed a Dot, found a micro-USB cable and plugged it in. The same sentences on step 0 cost them nothing.

The advice now lives in **one place**, read by both the panel and the throw. It is text somebody acts on, and two copies of instructions drift in the way that leaves half your users following the older half. The Connect button is disabled while it applies, since `navigator.usb` is undefined and the click can only fail.

## The test guards the advice, not the detection

Detection is one property read. The *advice* is what sends somebody somewhere:

- **The ingress branch must never offer `localhost:8768`.** `_ingress_only_middleware` rejects anything that is not the Supervisor gateway, so that is a 403 rather than a workaround — and it is the shape of the original #169 report. **Verified by reintroducing that exact bug and watching three checks fail.**
- **It must name Home Assistant's origin, not the controller's.** Chrome's allowlist matches scheme, host and port exactly, so somebody who allowlisted the standalone dashboard has an entry that silently covers nothing here.

Both functions are lifted from source rather than stubbed, so the test drives the real branch.

## Not attempted, deliberately

A downloadable or `file://` provisioner. Whether Chrome treats `file://` as a secure context is genuinely unsettled — and even where it does, the wizard is not self-contained: it fetches firmware, mints TLS credentials and creates the device row, which from a null origin is a CORS and auth surface on the endpoints that issue device credentials. A localhost helper *would* work, and is a signed per-platform binary serving one page for a step run once per device.

**The flag and HTTPS remain the answers.** This just says so before it costs anyone an evening.

https://claude.ai/code/session_012zuK5VZZut5EZm3jf3sPxk

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -83,6 +83,7 @@ jobs:
           node controller/tests/boot_target.test.mjs
           node controller/tests/pm_verdict.test.mjs
           node controller/tests/wizard_log_class.test.mjs
+          node controller/tests/web_usb_blocked.test.mjs
           node controller/tests/emos_md5.test.mjs
           node controller/tests/boot_image_length.test.mjs
 
--- a/controller/static/dashboard.jsx
+++ b/controller/static/dashboard.jsx
@@ -21,6 +21,39 @@ function isIngress() {
   return document.baseURI.includes('/hassio_ingress/');
 }
 
+// Why WebUSB is unavailable, or null when it is fine.
+//
+// WebUSB needs a secure context and no amount of code changes that, so this
+// exists only to say so BEFORE somebody unboxes a device — it used to surface
+// at the first click of Connect, with a cable already in their hand.
+//
+// One copy, read by both the wizard's pre-flight panel and requestDevice's
+// throw. A second copy would drift, and this is text somebody acts on.
+//
+// The origin is named because Chrome's allowlist matches scheme, host and port
+// exactly, and the add-on's page is served from HOME ASSISTANT's origin rather
+// than the controller's — so an entry added for the standalone dashboard does
+// nothing here and gives no clue why (#169).
+function webUsbBlocked() {
+  if (navigator.usb) return null;
+  const origin = window.location.origin;
+  return {
+    origin,
+    why: `WebUSB needs a secure context, and this page is on ${origin}.`,
+    // Under the add-on there is no localhost route to offer:
+    // _ingress_only_middleware rejects anything that is not the Supervisor
+    // gateway, so suggesting a direct port would send the user to a 403.
+    fix: isIngress()
+      ? `Serve Home Assistant over HTTPS, or add exactly ${origin} to `
+        + `chrome://flags/#unsafely-treat-insecure-origin-as-secure and relaunch `
+        + `the browser. An allowlist entry for the controller's own address does `
+        + `not cover this one.`
+      : `Open the dashboard at http://localhost:8768, or add exactly ${origin} to `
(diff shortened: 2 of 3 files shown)


wilbowes:
Closed by GitHub when its base branch was deleted on #471's merge, not by anyone. Reopened as #474 with the same commits rebased onto `main`.

PR #502, opened 2026-09-11
Run emOS on FireOS 6: system-as-root, our own combo-chip bring-up, supplicant and DHCP

wilbowes:
**emOS now runs on FireOS 6 and gets itself onto the network, with no Amazon binary in the path.**

Measured on a spare Echo unlocked with amonet v2.0.0: cold boot to internet — chip powered by our own ioctls, associated and WPA2-handshaked by our own supplicant, addressed by busybox udhcpc.

### What was actually wrong

- **FireOS 6 is system-as-root.** The partition's root *is* the Android root, with the real tree nested in `system/`, so every absolute `/system/...` path in `init.c` was one directory short. The mount succeeds either way, which is why it presented as a healthy boot with a cycling console banner rather than a failure. Fixed with a bind mount, not a path prefix — `vendor` and `etc` inside such a partition are *absolute* symlinks to `/system/...` and point at themselves once it is mounted there.
- **FireOS 6's networking binaries cannot be used at all.** `wmt_loader` exits 255 where FireOS 5's exits 0; `wpa_supplicant` and `dhcpcd` abort before `main()` on `open /dev/binder -> -1`. Rather than give emOS a property service and a binder node to keep Amazon's tools happy — which makes their userspace *more* load-bearing — emOS replaces them.

### What it does instead

- Brings the combo chip up from `init` against MediaTek's GPL driver interface, dropping `wmt_loader` **and** `wmt_launcher` on both kernels.
- Ships its own `wpa_supplicant` (hostap 2.10, static ARM32, nl80211, internal crypto, ~1MB) and uses busybox `udhcpc` on FireOS 6.
- Keeps its WiFi config in `/data/emos/wpa.conf`, out of reach of a FireOS 6 install on the other slot rewriting Android's file — which previously discarded the whole network block and left a device up, silent and needing a cable.
- Treats *no* credentials as a wait rather than a failure, which is the normal state while the wizard is provisioning over the console.
- Adds `em-wifi`: scan, pick, password, from the console.

### Not in this PR

The provisioning wizard **cannot install a FireOS 6 image yet** — the emOS release publishes only `init`, and the supplicant and `wpa_cli` need the same road. The packer half is done. A locally built image installs today.

### Also settled on hardware

FireOS 5 does **not** boot under amonet v2.0.0, and neither does emOS on the FireOS 5 kernel — so #500's warning is correct and understated. And WPA3 is one blocking layer rather than three: PMF turns out **not** to be blocked (the driver advertises BIP), leaving only SAE, which needs a kernel build.

Full reasoning, including the wrong turns, is in the JOURNAL entry for 2026-09-12.

https://claude.ai/code/session_01FuQdKPpaoa8MEaZ5hjNcXD


eeble:
Happy to help test this. I have an echo dot 2nd gen running amonet v2.0.0 simply because I found that project about 90 minutes before this one :smiling_face_with_tear:

Issue #541, opened 2026-09-15
Detect the board at startup, and resolve every hardware hook by name


wilbowes:
The firmware is written for one board. Three people are now working on others — radar (#535), Echo Input (#503), checkers (#36), crown (#358) — and each of them currently means editing the same constants in the same four files. This proposes the seam they should build against instead: the binary detects the board at startup and resolves every hardware hook by name.

## What already exists

More than it looks. `pkg/` carries four real interfaces — `led.Controller`, `buttons.Controller`, `mic.Microphone`, `speaker.Speaker` — and the server is written against them, not against the drivers. Capability announcement also already exists and is already conditional: `ambient_light` is announced only when the sensor actually reads.

So this is not a rewrite. Three things are missing.

## 1. Nothing selects a board

`cmd/server.go` calls concrete constructors directly:

```go
buttonController, err := internalbuttons.NewButtonController()
microphone, err := mic.NewMicrophone()
pcmSpeaker, err := speaker.NewPcmSpeaker(...)
```

One board is compiled in because it is the only one written. There is no registry and no detection.

## 2. Topology is hardcoded inside each binding

| Binding | Hardcoded | Should resolve by |
|---|---|---|
| `mic/pcm_microphone.go` | `cardNr = 0`, `deviceNr = 24`, 9ch S24_3LE | ALSA card/device **name** |
| `buttons/evdev_controller.go` | `/dev/input/event1`, `/dev/input/event2` | `EVIOCGNAME` |
| `led/i2c_controller.go` | sysfs paths containing `0-003f` | i2c device name |
| `als/als.go` | — resolves `tsl2540` by driver name | **already correct** |

`als.resolve()` is the model to copy, and its comment already states the rule: an address is an enumeration accident, hardcoding it works here and silently reads nothing on a device that enumerated differently.

Buttons are the live hazard. `event2` is the volume button on biscuit and the **touchscreen** on checkers — `evdev.Open` succeeds, nothing errors, and the buttons are simply dead. @imduffy15 found that on real hardware in #36.

## 3. Bindings cannot build off-target

`//go:build server` sits on the ALSA files, so a board package cannot be unit-tested on a CI runner while the descriptor (*what* to open) and the driver (*how* to open it) live in the same file. Splitting them makes board definitions testable without hardware.

## Proposal

**`pkg/board`.** A `Board` is identity, descriptors, and constructors. Boards self-register in `init()`; the registry picks one at startup.

[...]

PR #542, opened 2026-09-15
Read the serial from idme, where the hardware keeps it

wilbowes:
An emOS device on amonet 2.x reports no serial. The fleet is keyed on it, so every such device registers as `unknown-device` and they collide with each other — one identity, each overwriting the other's rows, and nothing reporting it.

## The cause is length

Measured on the spare (`G090LF11752215LE`) today:

| | bytes |
|---|---|
| LK prefix | 421 |
| stock image cmdline | 70 |
| LK suffix (holds `androidboot.serialno`) | 344 |
| **stock total** | **835** — fits |
| emOS image cmdline | 385 |
| **emOS total** | **1150** |

`androidboot.serialno=` starts at byte **1040**. FireOS 6 runs the 32-bit kernel — the reason we ship `init32` — where `COMMAND_LINE_SIZE` is 1024, so it is truncated away before the kernel sees it. FireOS 5 boots aarch64, where the limit is 2048 and the same string fits, which is why this looked specific to amonet 2.x.

Both parsers were correct the whole time. There was nothing to find.

## The fix is a better source

`/proc/idme/serial` is Amazon's ID Manager, exported by their kernel driver and world-readable — the hardware value, needing no property service and no bootloader argument.

| device | `/proc/idme/serial` | |
|---|---|---|
| v1, FireOS 5, Android | `G090LF1180571VVV` | identical to `getprop ro.serialno` |
| v2, FireOS 6, TWRP | `G090LF11752215LE` | correct where the cmdline is empty |

Both halves — `GetSerialNo()` in the firmware and `serialno()` in emOS's init — now try idme first and keep their existing sources as fallbacks, so nothing regresses where those already worked.

Trimming the cmdline was the alternative and is worse: it buys bytes by dropping FireOS arguments (`lowmemorykiller.*`, `veritykeyid`) whose only proof is a boot, and it leaves the serial hostage to a budget the next cmdline addition silently blows again.

A corrupt value is refused rather than passed on. This is an identifier the controller stores, logs and keys rows on, so a plausible-looking wrong one is worse than none.

## Tests

`emos/init/serialcheck.c` is the sixth off-target check (22 cases), and it earned its place before it was wired in — it caught two real bugs in the code it tests:

- `serial_copy` left partial output on rejection, handing back stale stack bytes that read as a perfectly valid serial
- `serial_from_cmdline` matched a longer argument merely **ending** in our key, the same trap `cmdlinecheck.c` already guards for `emos.system=`

Plus `device/internal/client/serial_test.go` for the Go half. All six off-target checks pass, both inits cross-compile for aarch64 and armv7a, and `go vet` and the device suite are clean (`internal/aec` fails locally on the known pointer-identity flake, untouched here).

**Not yet run on a booted emOS device** — verified against both devices over adb, and in TWRP rather than under emOS.

## Aside for #541

`/proc/idme` also carries `board_id`, `product_name`, `productid` and `device_type_id`, plus per-unit `alscal` and `miccal.0`–`miccal.6`. Worth knowing there: `/proc/device-tree/model` reads `MT8163` on both Dots — the SoC, not the board — so it cannot discriminate between boards on the same chip, while `device_type_id` (`A3S5BH2HU6VAYF` here) can.

https://claude.ai/code/session_01PWxgZMiimBJHHh5XgDn6oF


bogez:
Hit this on a fresh v2 unlock today, in case another data point helps.

Setup: Echo Dot 2nd Gen, amonet v2.0.0, FireOS 6574.1, controller 2.24.0-ea.7 (HA add-on), emOS v0.7, firmware v2.15.0.

Everything in the wizard went fine until the last step (Wi-Fi). It connected, then timed out saying the device didn't register. The controller log showed it was connecting, just as unknown-device:

[control] unknown-device: token presented but none on record. Treating as an unregistered device; it will need approval.

So I ended up with two pending entries: one under the real serial (which never connected) and unknown-device (which did). I approved unknown-device by hand and it's working. Assist commands go through fine.

One thing that might be worth handling until the fix ships: the wizard's last step can't succeed on these devices, and there's no hint to approve unknown-device manually. It took me a while to figure out.

(Also seeing [wilbowes/EchoMuse#546](https://github.com/wilbowes/EchoMuse/issues/546), where speaker switches 236/239 read Off. Looking forward to the next firmware for both!)

wilbowes:
Both should be fixed and are pending release. Hang in there, it's coming!

PR #543, opened 2026-09-15
Give the wake sensitivity slider a precise end that can actually fire

scragnog:
The Sensitivity slider's most precise notch set a wake threshold of exactly `1.0`, which nothing can clear — openwakeword's score is a sigmoid that saturates below 1.0 and both scorers test `score >= threshold`. The strictest setting was a wake word that could never fire, and it presents as a device that has stopped responding rather than as a value set too high.

It is an off-by-one. The mapping is `1.0 - 0.1*(v-1)` against the `0.1–0.9` range its own comment documents, so every notch sits one step too strict and the eager end stops at 0.2 instead of 0.1.

## What changed

Range is now **0.1–0.975 at 0.025 per step**, driven through the shared `Slider` so the threshold appears in the readout the way Barge threshold and Arbitration window already do.

0.975 rather than the documented 0.9 because 0.9 was already the strictest *usable* setting — capping there leaves anyone sitting at 0.9 (because 1.0 was dead) with nowhere to go, and that band is where the precision/recall tradeoff actually lives.

Stored config is the threshold, not the slider index, so existing installs keep their exact value and only the handle moves. The 0.5 default lands on a notch.

## Evidence

18,021 scored frames, three Echo Dot Gen 2s:

| threshold | frames at or above |
|---|---|
| 0.90 | 488 |
| 0.95 | 343 |
| 0.98 | 178 |
| 0.99 | 104 |
| **1.00** | **0** |

Peak observed score across the set was 0.999.

## Tests

`controller/tests/wake_sensitivity.test.mjs`, added to the `dashboard-tests` job. It lifts the range constants and `reflectT` from source in the style of `number_field.test.mjs`, then asserts no reachable threshold is `>= 1.0`, the precise end reaches past 0.9, `reflectT` is its own inverse, the range is a whole number of steps, and the default lands on a notch. It also pins the `>=` in `em_controller.py` and `shadow.go`, since that comparison is what makes 1.0 unreachable.

Confirmed it fails against the old mapping before passing against the new one.

## Tested on

Three Echo Dot Gen 2s, firmware v2.15.0, `owwOnDevice: "on"`, controller 2.23.1 on a native macOS install. Running at 0.975 and waking reliably.

## One thing worth your call

The slider still runs Precise → Eager left to right, so the displayed number *increases* as the handle moves left. Keeping the existing direction avoids changing muscle memory, but it is the opposite convention to Barge threshold, which is also a threshold and runs low → high. Happy to flip it if you would rather the two matched.


wilbowes:
Good find, and the diagnosis is exactly right. I confirmed `sensitivityToThreshold(1)` returns exactly `1.0` on main, and both scorers compare with `>=` (`em_controller.py:2976,2982` and `shadow.go:370`). The strictest notch was a wake word that could never fire.

The new mapping checks out: 36 notches from 0.1 to 0.975, `reflectT` is its own inverse, 35 whole steps, and the 0.5 default lands on a notch. Every `Slider` prop you pass exists, and omitting `disabled` matches the other 16 sliders in that form. Test passes.

**One thing before this goes in.** `1.0` was reachable on the old slider, so there may be fielded devices stored at it. A range input clamps an out-of-range value to `min`, so such a device now renders:

```
stored=1.0  →  track=0.075  →  clamped=0.1  →  readout=0.975
```

The dashboard shows a plausible value for a device that cannot wake. Your change doesn't introduce that, and it's the natural place to close it — clamping on read, or writing the clamped value back, does it in a line.

Minor: the test hardcodes `DEFAULT = 0.5` rather than reading it from `em_db.py`, so a default later moved off the grid would still pass. Same shape as the bug you're fixing.

Happy to take it with the clamp added.

scragnog:
Both done in `f18faca`.

**The clamp.** `wakeTrackFor` pins the handle, and the readout now renders the **stored** threshold rather than the clamped track position — so a device at 1.0 reads `1.000`, not `0.975`. It stays wrong on screen only while it is wrong on the device, and one drag writes a value on the track.

I went with that rather than writing the clamped value back, because the write-back is unreliable here: `set()` is gated on the owning section and `owwThreshold` belongs to `wakeword`, so the write is silently dropped for precisely the devices inheriting the fleet value. A migration that skips the inherited case seemed worse than none — it would leave the stored value and the display disagreeing, which is the same failure in a quieter form. Happy to add it on top if you'd rather have both.

**The default.** Now read from `em_db.py`. Confirmed it fails if the default moves off the grid (tried `0.52`), which it did not before. Also covers a stored `1.0` pinning to the track, and pins the readout's source shape.

**One thing I left for you.** A fielded device at 1.0 still cannot wake until somebody opens the dashboard and drags. Clamping server-side where `oww_threshold` is read (`em_controller.py:3511`, `em_api.py:1099`) would repair it without that, but I did not want to do it unasked: the obvious ceiling is this slider's 0.975, and that would quietly loosen a deliberate `0.99` set through the API — which is reachable, 104 of the 18,021 frames cleared it. That is a policy call about whether the controller should refuse a threshold it can never satisfy, and it is yours rather than mine. Say the word and it is a separate PR.

— Rob

wilbowes:
Merged, thanks Rob. Sorry it sat for three days after you'd done everything asked.

Your open question was already settled by #549, which caps the threshold at 0.975 whenever it's saved. A 1.0 stored before that stays until the next save, so we'll clear those with a one-off migration. On direction: we're keeping yours (Precise → Eager) and flipping the barge-in slider to match, so both threshold sliders read the same way.

PR #547, opened 2026-09-16
feat(forge): show build progress per stage and collapse tqdm redraws in the console

### LINKED ISSUE #266: Forge: show training progress against the wake word, and tame the log
A build runs for one to two hours on a GPU and overnight on CPU, and the only signal is a log tail. You cannot tell from the UI whether it is a third of the way through or stuck.

Two halves, and the first is nearly free.

### Progress per stage

The three build stages each have a countable signal, and two of them need no new plumbing:

- **generate** — `clips_train` is already in `/api/state`, and `n_samples` is already on the card. Progress is the ratio. This is the long stage and the one worth showing.
- **augment** — four `.npy` feature files are written to the work directory. Coarse but honest: 0–4.
- **train** — `train.py` prints step numbers and `steps` is in the config, so it needs a parse of the job log rather than a disk check.

That suggests the stepper on each card grows a fill rather than a bulb: the stage that is running shows its own ratio, the ones behind it stay solid. It also keeps the existing "state is derived from disk on every poll" property for two stages out of three, which is what makes it survive a container restart.

### The log itself

`train.py` and the piper generator both use tqdm, which redraws with carriage returns. The console appends raw text into a `<pre>`, so a progress bar arrives as thousands of near-identical lines and buries anything useful. Collapsing `\r` segments to the last one per line would fix that on its own, and is worth doing whether or not the progress bar happens — the same treatment makes the 100KB client-side log buffer hold far more of the real output.

Worth checking while in there: whether the log ring should keep the last N *meaningful* lines rather than the last N bytes, for the same reason.

### Note

The one-job lock is per web-UI process, so a `docker compose run` started alongside it is unguarded and would make any progress figure wrong. Not a reason to avoid this, but the display should read from the wake word's own directory rather than assuming the running job owns it.

costajohnt:
Progress per stage on the wake word card while a build runs, and a console that no longer drowns in tqdm redraws.

**Progress.** `/api/state` gains `progress: {stage, done, total}` on the word being built, null otherwise. The stage comes from forge.py's own `=== step: X ===` lines in the job log; the directory alone cannot tell a clip top-up on a word that already has features from a finished generate stage. The ratio for generate and augment is counted from the word's directory on every poll, like the rest of the state, so it is right whichever process is writing there: generate counts all four clip dirs against `2 × (n_samples + n_samples_val)`, capped per dir so neither `positive_train` alone nor a Google TTS overshoot reads as 100% with the negatives still to run; augment counts feature files finished (a file exists from the moment train.py opens its memmap, so present minus one). Train has nothing on disk until the `.onnx` lands, so it is parsed from train.py's log: the `Training: ... N/M [` tqdm redraws plus the `Starting training sequence N` markers, folded across auto_train's three sequences (`steps`, `steps/10`, `steps/10`) into one `done / total` so the bar fills once instead of hitting 100% and restarting twice. A bar that starts over also counts as a new sequence, so the fill does not depend on the markers (they are `logging.info` lines that only appear because train.py imports piper's `generate_samples.py`, which configures logging at import). Once a bar has been seen its own total replaces the config's `steps`, since the config can be edited under a running build.

The fold is incremental on the `Job` and bounded to 1 MiB per poll: nobody may have had the page open since the build started, and a CPU run's log is tens of MB by morning, which folded in one go would hold the event loop for seconds. If progress cannot be read (log removed under the job, a hand-edited `steps: 50k`) the field is null, the UI falls back to the old pulsing bulb, the poll itself is unaffected, and one `[forge-ui] no progress for ...` line goes into the job's console.

The stepper's running step shows a fill bar with an `N / M` caption instead of the pulsing bulb; other jobs on a word (Google TTS, extra voices) keep the bulb.

**Log.** `pollLog` keeps only the last `\r` segment of each line, carrying the unfinished tail across polls so a live bar keeps updating in place and a line split across two 64KB reads is collapsed whole. The 100KB client buffer therefore holds real output rather than a few seconds of bar redraws. I left the byte ring as is; with the collapse it seemed to cover the "last N meaningful lines" question, happy to change it if not.

Known coarseness: the train number sits at 0 while train.py loads data and at 100% through checkpoint merge and export; augment is four files, so it moves four times.

**Tested on:** `python -m pytest oww_forge/tests/` (15 new tests in `test_forge_progress.py`, pure module, no training stack; the tqdm fixtures are verbatim tqdm 4.67 output in both ASCII and unicode bar styles) and `node oww_forge/tests/console_log.test.mjs` (lifts the console fold out of index.html the way `controller/tests/*.test.mjs` do, checks every split point of a synthetic tqdm log; added to the dashboard-tests CI job). `_wakewords_state` was also exercised against a fake job writing a synthetic log through every stage, plus a log deleted mid-job and a config edited to `steps: 50k`. Not run against a live build: the tqdm format, the sequence markers and the memmap behaviour were read from the pinned openWakeWord (`368c0371`) and piper-sample-generator (`v2.0.0`) sources rather than observed, so if a real log differs from `Training:  25%|##4       | 1234/5000.0 [` the regex in `forge_progress.py` is the place to look.

The four feature file names forge.py's heal path already listed move to a `FEATURE_FILES` constant both files share; no behaviour change there.

Fixes #266


wilbowes:
Merged, thanks. This is a careful piece of work: folding the three training sequences into one bar, capping how much of the log each poll reads so an overnight CPU log can't stall the page, and the fallback to the old pulsing bulb when progress can't be read. I resolved a one-line `ci.yml` clash on your branch before merging.

As you said, it hasn't met a live build yet, so if the first real one shows a stuck or wrong bar, the regex in `forge_progress.py` is where we'll look. It ships in the next forge image.

Issue #295, opened 2026-08-23
LED ring brightness: make it a setting, and expose it to Home Assistant


wilbowes:
**There is no brightness setting today.** The only LED config keys are `ledScene`, `ledListenColor` and `ledThinkColor` — checked, not assumed. So this is new work rather than plumbing something existing through to HA.

That matters for ordering: brightness has to exist as a controllable value before anything — Home Assistant, an automation, or the ambient light sensor (#295) — can drive it.

## The user-facing problem

A twelve-LED ring at full output is a lamp. In a dark bedroom at 3am it is the brightest thing in the room, and the only way to stop it today is to pick a scene that happens to be dimmer.

## Scope

- **`ledBrightness` as a device config key**, in the `ring` section, scoped per device like the rest of that section. Someone will want the bedroom dim and the kitchen bright, and the fleet/device split already handles that.
- **Applied at render**, not per scene. It scales whatever the priority system is currently drawing, so it works for the listening ring, the volume arc, the mute indicator and the playback meter without each one knowing about it.
- **Exposed to HA** so it can be automated — a number entity is the honest shape, since it is a continuous value rather than a light with colour.

## Two things to get right

**Do not let brightness reach zero and become a stealth mute indicator.** The red mute ring is a privacy signal, and a user who has dimmed the ring to nothing has silently disabled the only visible evidence that the microphone is off. Either floor the brightness above zero, or exempt the mute state from scaling. This is the kind of interaction that looks like a rounding detail and is actually the whole point of the indicator.

**Gamma, not a linear multiply.** Perceived brightness is nonlinear, so scaling raw channel values by 0.5 does not look like half. The meter already carries `meterGamma` for related reasons; whatever curve is used should be stated in the code rather than discovered by someone matching two devices by eye.

## Related

- #296 — driving this automatically from the ambient light sensor, which depends on this landing first
- #66 — letting HA drive the ring's content when idle; separate feature, same entity surface

Prateekathub:
Hi! I would love to handle this as my first contribution. Could you assign it to me?

wilbowes:
@Prateekathub yes please, it's yours, and sorry for the slow reply. The scope in the description is the plan: a `ledBrightness` config key in the `ring` section first, then exposing it to Home Assistant. Ask here if anything's unclear.

It also ties in with #66, which will let Home Assistant drive the idle ring, so a brightness setting that applies to every animation is exactly the right foundation.

— Team EchoMuse (powered by Claude)

Issue #66, opened 2026-08-02
Let Home Assistant drive the LED ring when the device is idle


wilbowes:
The ring is a 12-LED addressable display sitting idle in every room of the house. Right now it only ever shows EchoMuse's own state — listening, thinking, playback meter, mute, volume.

Idea: expose it to HA while idle, so it can carry ambient household information — which bins go out tonight, flashing on a doorbell press, a colour for weather or an alarm state.

**Constraints that already exist and should be respected:**
- Device-local animation is already there (`led_anim`: `solid|spin|rotate|pulse|meter|off`, colours, `periodMs`, `ttlSec`) — an HA-driven scene should ride that rather than streaming frames, so WiFi jitter cannot judder it.
- `ttlSec` is a dead-man switch. An HA-set pattern must expire on its own if HA stops talking, or a stale ring becomes permanent.
- **Turn states must outrank it.** Listening, thinking and mute already have a priority order in `server.go`; HA gets the ring only when idle, and loses it the instant a turn starts.
- Mute ring (red) and volume arc (cyan) are device-sovereign and must stay that way.

**Probable shape:** a `light` entity per device, or a small set of HA-callable scenes. A `light` gets HA's colour picker and automation support for free, but it needs to be clear it is an *indicator*, not room lighting.

linuxstony:
Great idea! - e.g. red turning circle when alarm is armed.

larribald:
Yes! I came here looking to see if anyone else had raised exactly this idea. I was thinking of a brief flash to notify the start of an automation, but your examples are even better

wilbowes:
This now has a first concrete use. #552 adds a Home Assistant switch that turns an Echo's wake word off, with no ring indicator of its own. With this issue, the same automation that turns the wake word off could show whatever the user wants on the idle ring, as @electroflame suggested in #286.

We'd like this. The constraints in the description (ride `led_anim`, keep `ttlSec` as a dead-man switch) still stand.

— Team EchoMuse (powered by Claude)

evy0311:
I started working on this exact feature before I realized there was this open issue for it. @wilbowes , can you please assign it to me? I have it working locally with only changes to the controller. 

Edit:
[...]

PR #552, opened 2026-09-16
Report the Mute status to HA, provide a software mute for it to use

### LINKED ISSUE #286: Virtual Mute Switch
Exposing a virtual mute switch to Home Assistant would be very useful.  I try to mute satellites if the TV is currently playing (and unmute when paused/stopped) but I can't do that with EchoMuse as there's no exposed switch.

Lighting up the full light ring for this would be too much (no need for a giant glowing red ring while watching TV), my suggestion would be to only light up the actual red mute-button LED, which would also differentiate it from a full, you-need-to-press-the-button mute.  Full-ring, have to press the button; just the button led, virtual toggle, controllable by HomeAssistant.

As a bonus it'd be nice to be able to control the LED brightness for this, so the mute button can be dimmed/brightened depending on room lighting (to make it less distracting in a darkened room, for instance).

### LINKED ISSUE #438: Report mute state to Home Assistant, so an automation can avoid asking a muted Echo
**Home Assistant cannot tell whether an Echo's microphone is muted, so an automation cannot avoid asking a question of a device that will never answer.**

We advertise an event entity, a media player and a sensor. None of them carries the mute. The media player's `muted` field is hardcoded `False` and is about volume anyway, so there is nothing to read even by accident.

### Why this is sharper than it looks

**A muted device runs the turn and captures nothing, deliberately.** The device rejects every `mic_start` while muted and the ADC is muted in hardware, so the turn streams silence, gives up at its cap, and HA ends the run with no answer. Refusing the turn controller-side is the change that looks safer and is worse — `assist_satellite.ask_question` awaits its answer future with **no timeout**, so a satellite that refuses by staying silent hangs the caller's script for good.

That is the case worth fixing from the other end: if HA can read the mute, an automation can decline to ask rather than discovering it by hanging. It came up in #423, where a question is asked and no answer arrives — a muted Echo produces exactly that signature and is worth ruling out first.

### Shape

- **Read-only, and that is the point.** The hard mute's value is that software cannot undo it: it cuts the microphone at the codec and the device refuses every `mic_start` while it holds. A writable entity here would create a remote unmute, which is the one thing the mute exists to make impossible. The writable half is #286, as a *soft* mute that is a different thing with a different ring colour.
- **Announced per device**, gated on the `mic` capability, like every other entity — and it has to survive the entity list being a one-shot at `ListEntities`, so a capability arriving late needs the connection bounced rather than the entity silently missing for the life of the connection.
- **It should distinguish hard from soft** once #286 exists, or an automation cannot tell a mute it can clear from one it cannot.

Pairs with #286 (HA-controlled soft mute) — that issue is the write side, this is the read side, and the soft mute is much less useful without something that reports state.

DaSonOfPoseidon:
Two entities per device, gated on `mic` like the others. Controller-only; no firmware or protocol change.

- **Microphone Muted** (`binary_sensor`, key 4, read-only) — the button mute as the device reports it, so an automation can skip `ask_question` on a device that would never answer. Nothing accepts a command for it: the hard mute stays unclearable from software. Closes #438.
- **Soft Mute** (`switch`, key 5) — on, the wake word is ignored and the wake stream is stopped (`mic_stop`). HA-initiated `start_conversation`/`ask_question` still listen, as they do under the hard mute. The button wins in both directions; a reconnect re-report of the same hard state leaves it alone, so a Dot dropping off Wi-Fi mid-film comes back still muted. Closes #286.
- **Ring: a steady dim violet `(60,0,120)`** while the soft mute holds — the one hue nothing else on the ring uses (red is the button's, orange the link, cyan volume, white pending, and the scenes take green/blue/pink/rainbow). A state, so no TTL; painted from `leds_off`, so it returns by itself after any turn, announcement or alarm, and after a turn-end cue expires. The colour is one constant in `em_scenes` — a `ledSoftMuteColor` key beside `ledListenColor`, or HA's own pick via the `light` entity #66 proposes, would replace it there and nowhere else. Lighting only the mute-button LED needs a `mute_led` control message plus a capability (bare GPIO only `mute.go` drives, on/off only, so the brightness ask belongs with #295) — left as a follow-up because I have no Dot to test firmware on.

Decisions are in `em_softmute.py` (pure, tested). `Device.mic_start` is the one gate for the wake stream, so the nine restart sites need no change; `mic_start_turn` is untouched.

**Tested on:** the published image with this source overlaid (CI's import job, run locally) + a fake device on `/control` + `aioesphomeapi` as HA: entities advertised, initial states on subscribe, sensor follows `mute_state`, switch sends `mic_stop`/`mic_start` and paints/clears the ring, button clears the switch, reconnect keeps it — 25 checks. Unit suite 1113 passed. **No rooted Dot** — nothing device-side exercised.

Resolves #286
Resolves #438


wilbowes:
Thanks for this. The mute sensor is exactly right, and read-only is the correct call.

One change before we merge: we'd like the soft mute presented as **"Wake word"**, a switch that's on by default. As written it ignores the wake word while `ask_question` and `start_conversation` still listen, which is the behaviour we want, but "mute" suggests the microphone is off, and on the Dot 2 only the button does that. It also matches #286's use: turn the wake word off while the TV is on. So:

- Rename the switch to "Wake word" (on means listening), and update the FAQ, quickstart and LED-ring docs to match.
- Leave it alone when the mute button is pressed. The button mute and this switch should stay fully independent. That reverses what the plan in #286 said, so sorry for moving it on you.

Everything else, the sensor, the violet ring and `em_softmute.py` with its tests, can stay as it is. Thanks for testing it so thoroughly without a Dot.

wilbowes:
One more change, on the ring: please drop the violet, so turning the wake word off leaves the ring as it is.

@electroflame asked in #286 for no ring indicator, and the better home for one is #66, which lets Home Assistant drive the idle ring. An automation that turns the wake word off can then show whatever the user wants on the ring, or nothing. That keeps `em_scenes` free of a colour we'd otherwise have to make configurable later.

DaSonOfPoseidon:
The "Soft Mute" switch is now "Wake word detection"; it is independent of the mute button, and it no longer paints the ring.

Three changes from the previous commit:

- Renamed and inverted. em_softmute.py → em_wakeword.py; the HA entity is switch.wake_word_detection (on by default, on = listening). "Soft mute" implied the microphone was off, which it never was — an HA-initiated start_conversation/ask_question still listens. Not named plain "Wake word" because HA's ESPHome integration already puts a picker by that name on the same device page.
- The button no longer moves the switch. Pressing mute or unmute leaves the switch where HA put it. The one thing the button still forces is the stream: the device restarts its own wake stream on unmute, so an unmute with the switch off sends a mic_stop to take it back down — re-asserting the switch rather than changing it.
- No LED state. The violet idle ring, SOFT_MUTE_VIOLET, the scene's soft_mute_anim and the post-cue repaint are gone. The ring says nothing about this switch; HA driving the idle ring is #66.

Smaller notes: 
The wake-word picker's read side now reports active_wake_words=[] while the switch is off, so HA's dropdown shows the truth after a reconnect (it still declines writes, with the reason in the log). 
/api/devices exposes wake_word from the ESPHome server rather than live Device state, so it answers for an offline device too and reads null when there's no server.
Tests moved to test_wakeword.py; docs (FAQ, quickstart, configuration, led-ring-states) updated to match.

wilbowes (changes requested):
Thanks, this is what we asked for, and the name "Wake word detection" avoids the clash with HA's own picker.

One change, from something that landed after you wrote this: private listening (#602) is now the default, and there the Echo detects the wake word itself and sends `oww_wake` with a session. `_private_wake_turn` only checks `device.muted`, so with the switch off a wake still starts a turn. Could you add the switch next to that check, closing the session with `listen_close(session, "wake_off")`?

That needs a rebase onto main, which conflicts in `em_controller.py`. If you'd rather we did the rebase, say so and we'll push it to your branch.

Small ones: the stray blank lines in `test_scenes.py` and `led-ring-states.md`. Keys 4 and 5 stay yours; #620 will move to 6.

DaSonOfPoseidon:
I can handle the rebase and make the updates... I'll probably be able to get to it before the end of the week.

DaSonOfPoseidon:
jk it was easier than i expected it to be

PR #554, opened 2026-09-16
Support for emos on Echo 2 (radar)

vithurshanselvarajah:
### What

Per-board runtime and headers for emOS init. Adds the seam for a second board to slot in without touching `init.c`.

### How

| Layer | Change |
|-------|--------|
| `emos/init/boards/boards.h` | New. Runtime interface — `board_log_fn`, `struct board_node`, prototypes for `board_set_log`, `board_nodes`, `board_wifi_up`. |
| `emos/init/boards/boards_stubs.h` | New. `__attribute__((weak))` fallbacks so the off-target tools (`cmdlinecheck`, `ringsim`, etc.) link without `boards_*.c`. |
| `emos/init/boards/biscuit.h` | New. Every biscuit literal — `CACHE_PART`, `BOOT_PART`, `LED_*`, `biscuit_nodes[]`. |
| `emos/init/boards/boards_biscuit.c` | New. MediaTek `wmt_bringup` / `wmt_daemon` / `wmt_answer_patches` moved here. |
| `emos/init/init.c` | Calls into `board_nodes()` and `board_wifi_up()` instead of touching literals. Reads `cmdline_board()` from `/proc/cmdline` and logs the resolved id. |
| `emos/init/cmdlinecheck.c` | Covers `cmdline_board()` with the same token-boundary / non-printable / length-overflow cases as `cmdline_system_part()`. |
| `emos/build.sh` | Links `init.c` + `boards_biscuit.c` into the static init binary. |
| `controller/em_emos_build.py` | `_fdt_compatible()` reads root `compatible` from the reference's first DTB (~70 lines of stdlib, big-endian per FDT spec). `reference_board_id()` normalises vendor prefixes (`mediatek,mt8163-biscuit` → `biscuit`). `pack()` stamps `emos.board=`; `build_emos_image()` resolves explicit > DTB > `BOARD_DEFAULT`. |
| `emos/mkboot.py` | Mirror of the above for the standalone tool, with `EMOS_BOARD` env override. Pinned byte-for-byte to the controller's copy by the existing parity test. |
| `controller/tests/test_emos_build.py` | Eight new tests for `reference_board_id()`, dedup, priority order, non-printable rejection, and packer parity. CMDLINE fixture updated to carry the new stamp. |

### How a board gets picked

1. `board_id=` argument to `build_emos_image()` wins.
2. Otherwise `reference_board_id()` reads `compatible` from the reference's first DTB.
3. Otherwise `BOARD_DEFAULT = "biscuit"`.

The stamp lands at the end of the cmdline, after `ramoops` and `emos.system=`, with the same dedup rule (replaces rather than appends).

### What this does NOT do

- No runtime selection. `init.c` reads `cmdline_board()` and logs it; the actual board runtime comes from whichever `.c` was linked at build time. Today that's `boards_biscuit.c`. To support a second board, write `boards/<name>.h`, write `boards_<name>.c`, and pick which TU compiles.
- No registry in the controller. The packer stamps what it's told — a wrong stamp boots an init that has no matching TU, with a hard failure at wifi bring-up (`/dev/stpwmt: errno=19`).
- No fix to the pre-existing FDT-magic byte-order inconsistency in the existing packer search. The new parser matches that search; a separate change should flip the search to the canonical byte order and the parser along with it.

### Test plan

- `pytest controller/tests/test_emos_build.py` — 68 pass (8 new, 60 existing).
- `pytest controller/tests/test_emos_image_fields.py` — 4 pass.
- Off-target tools: `cmdlinecheck` exercises the new `cmdline_board()` parser; `ringsim`, `pwcheck`, `tmoutcheck`, `wpacheck`, `serialcheck` all link cleanly via the weak stubs.
- Hardware: the build produces a byte-for-byte-equivalent image to a pre-this-PR build for any reference where the DTB-sniff resolves to `"biscuit"` (the default). Boot trail gains one new line: `board id=biscuit`.

### Adding the next board

```
emos/init/boards/donut.h       # donut-specific constants and nodes table
emos/init/boards/boards_donut.c # donut wifi-up and node getter
emos/build.sh                   # BOARD_SRC=.../boards_donut.c
controller/em_emos_build.py     # one elif in reference_board_id() if the short name needs special-casing
```

No `init.c` changes.


vithurshanselvarajah:
@wilbowes If you get a chance. I wouldn't mind a quick review to make sure this code layout fits to your expectations. 

Considering emos is going to be the standard going forward. It felt relevant to do this before i started working on adding radar support.

wilbowes:
Thanks, will take a look. Just bear in mind we've only been building emOS for a week and still ironing out bugs - the foundations might shift a little as we go.

der-wahshun:
Hey, I've got an Echo 2 (`radar_puffin`) here as well and have been testing the radar branch on it.

I managed to build the current branch successfully with `EMOS_BOARD=radar` and `EMOS_SYSTEM_PART=13`. My Fire OS p13 is still intact (`NS65741/8138`) and has `wmt_loader`, `wmt_launcher`, `WIFI_RAM_CODE_8163` and `WMT_SOC.cfg`.

I had a look at the original Fire OS init files too. They start `wmt_loader` followed by `wmt_launcher -p /vendor/firmware/`, and the device exposes `/dev/wmtdetect`, `/dev/stpwmt` and `/dev/wmtWifi`.

Since Wi-Fi seems to be the main blocker right now, I'm happy to use this device for testing. I was thinking of trying the original Fire OS WMT bring-up sequence instead of the current `SET_STP_MODE` path.

Let me know if you're already working on that or if you'd like me to test anything specific.

PR #565, opened 2026-09-17
Keep the AnnounceRequest line in support bundles, drop turn traces by name

### LINKED ISSUE #507: The support bundle drops every AnnounceRequest line, because the redactor keys on text=
`em_support._LOG_DROP` contains `text=`, and the announcement log line is:

```python
log.info(
    f"[{self._log_name}] AnnounceRequest: media_id={msg.media_id!r} "
    f"text={msg.text!r} start_conversation={msg.start_conversation}"
)
```

So every support bundle drops the only line that records an announcement arriving — and with it the only line that distinguishes `assist_satellite.ask_question` and `start_conversation` from a plain announce, since `start_conversation` is a field on that message and nothing else logs it.

Found while diagnosing #506. The bundle showed announcement audio playing (`duck held (depth 2)`, a buffered `post_turn_play`) with no announcement ever received, which is not a state the code can reach. It cost an hour before the redactor turned out to be the cause.

The marker was meant for turn traces, which quote transcripts verbatim. This line's `text=` is a TTS string HA sent us, and it is already covered: `{msg.text!r}` is quoted, so `_QUOTED` replaces it with `<redacted>` and the rest of the line survives.

## Fix

Drop `text=` from `_LOG_DROP` and let `_QUOTED` handle it, or exempt this line explicitly. A test asserting an `AnnounceRequest` line survives sanitising with its `start_conversation` field intact would keep it fixed.

Worth checking the other three markers against the same question: is this marker catching lines whose payload the quote and URL rules already cover?

## Also affected

#373 is an announcement fault, so the bundles on it are missing this line too.

costajohnt:
`_LOG_DROP` carried `text=` for the `[TURN]` trace, and the AnnounceRequest line renders `text=` too, so every bundle lost it along with `start_conversation`. The marker is now `[TURN]`: the trace still goes whole, and the AnnounceRequest line falls through to `_QUOTED`, which turns `text='Dinner is ready'` into `text=<redacted>` and leaves `media_id` and `start_conversation` in place.

I went with naming the line rather than deleting the marker, since deleting it would also un-drop the turn trace and rest transcripts on the regex, which the comment above `_LOG_DROP` says not to do.

On the other three markers: `STT result` matches the `STT result: {text!r}` line in `em_esphome.py` and `Utterance saved` matches the recording line, both transcript sources and both still needed. `stt_text` matches no rendered log line in the controller (only code and a dict key), so it is inert; left in place.

Known limitation, left for a follow-up: `_QUOTED` stops at the first matching quote, so a `repr()` value holding both quote kinds (`text='it\'s "late"'`) redacts as `<redacted>s <redacted>'`. Only the AnnounceRequest line is affected and only when the text mixes quote kinds.

**Tested:** `cd controller && python -m pytest tests/` (1102 passed, 3 skipped on Python 3.12; two new tests in `test_support.py`, the first fails on `main` with the AnnounceRequest line missing entirely, the second with the leaked fragment). pyflakes undefined-name check clean. No hardware involved; `em_support` is pure logic.

Closes #507


wilbowes:
Thanks, the fix is right: matching the turn trace by `[TURN]` instead of `text=` keeps the AnnounceRequest line, and `start_conversation` with it. I checked every log line that prints `text=`, and there are only those two.

One change before we merge. It's the limitation you flagged. Before this PR the whole AnnounceRequest line was dropped; now its text goes through `_QUOTED`, and announcement text is written by the user in their automations, which bundles must never carry. So the mixed-quote leak is new exposure, even if narrow. A pattern that understands Python's `repr` escapes closes it:

```python
_QUOTED = re.compile(r"'(?:[^'\\]|\\.)*'" r'|"(?:[^"\\]|\\.)*"')
```

Every case I tried, including your `text='it\'s "late"'`, redacts whole. A test for that case would pin it.

Optional, if you're in there: the spoken-dismissal line (`Spoken dismissal {text!r}` in `em_esphome.py`) logs a transcript and no drop marker catches it. It's already on main rather than something you introduced, but `Spoken dismissal` in `_LOG_DROP` would follow the rule that transcript lines go whole.

costajohnt:
Done, pushed as a second commit.

- `_QUOTED` now uses your pattern for both quote kinds, so an escaped inner quote no longer ends the match. `text='it\'s "late"'` redacts to `text=<redacted>`; test added.
- I kept the plain `'[^']*'` / `"[^"]*"` forms as trailing alternatives. Repr never needs them, but `_scrub` runs the same regex over device shell output and exception text, which is not repr, and there a value ending in a lone backslash (`'C:\temp\'`) would not close under the escape-aware form and would go out unredacted. With the fallback it redacts as before. The one shape neither form gets right is two adjacent values where the first ends in a backslash (`'foo\' 'bar'`); that was already wrong before this PR and no single regex sorts it out, so I left it.
- Took the optional one too: `Spoken dismissal` is in `_LOG_DROP`, with a test that the line is dropped whole and a neighbouring line survives.

`test_support.py`: 46 passed locally.

wilbowes:
Merged — thanks @costajohnt, both for the fix and for taking the escaped-quote pattern and the dismissal line further than asked. Support bundles keep the AnnounceRequest line now, and #507 is closed.

PR #604, opened 2026-09-21
Pin AsyncZeroconf to SERVER_IP's own interface, not every host interface

NicFragale:
zeroconf's default interface detection opens one socket per interface on the host, and on a multi-homed box that includes whatever else is attached — in this case a separate monitoring VLAN (Monitor30/40/50) that carries real IPv4 addresses and gets auto-detected like any other interface. A send on that socket can fail independently of the interfaces mDNS actually needs (measured: OSError: [Errno 19] No such device), and the exception surfaces inside the asyncio event loop rather than being contained to the interface that caused it — stalling the loop long enough to miss keepalive pings and disconnect the whole fleet.

SERVER_IP is already the one address every device is told to dial, so binding zeroconf to anything else would be a second kind of wrong answer.


wilbowes (changes requested):
Thanks for this. Advertising only on the address devices actually dial is right, and the fix is nicely small.

One change before we merge: if SERVER_IP isn't an address on this host (a DHCP change, a stale `.env`), `AsyncZeroconf(interfaces=[SERVER_IP])` now raises `OSError: [Errno 19] No such device` at startup and the controller exits. Could you catch that in both places, log that SERVER_IP isn't on any local interface, and fall back to the current `AsyncZeroconf()`?

Also worth a line in `docs/configuration.md`: with this, HA has to be on the same network as SERVER_IP to discover the satellites over mDNS.

